PR #2, opened 2019-12-23
add support for tolerations

AdheipSingh:
Adding Support for tolerations in order to run druid on tainted node.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ druid-operator$ kubectl apply -f examples/tiny-cluster.yaml
 
 # Debugging Problems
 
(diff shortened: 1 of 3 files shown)


AdheipSingh:
For Kubernetes 1.11, this CRD shall throw an error of this sought 
```
 ExternalDocs:(*apiextensions.ExternalDocumentation)(nil), Example:(*apiextensions.JSON)(nil)}: must only have "properties", "required" or "description" at the root if the status subresource is enabled
```
Hashing out the ```type: object``` shall work. 

Referenced Issued Here 
```
https://github.com/kubernetes-sigs/controller-tools/pull/72
```

PR #11, opened 2020-02-10
HPA support in Spec

AdheipSingh:
Fixes #11 HPA support in CR Spec

### Description

Goal is to make the operator support Horizontal Pod Autoscaling for scaling druid pods.Add support in the druid node spec, so any user of this operator can have the flexibility of specifying 

Add support in the druid node spec, so any user of this operator can have the flexibility of specifying HPA on which ever node, which feels as necessity.

The node replicas should first check if hpa is deployed and get min replica count from the HPA spec. This way in case HPA scales the pods, the operator does not reconcile to its  state. 

<hr>

This PR has:
- [ x] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
- [ x] added documentation for new or modified features or behaviours.

<hr>

 * `GetHPAReplicaCountOrDefault` - Func Added for getting the replica count in case HPA is used.

 * `Replicas: IntPointer(GetHPAReplicaCountOrDefault(sdk, types.NamespacedName{
				Name:      serviceName,
				Namespace: m.Namespace,
			}, nodeSpec.Replicas,` - Replicas to check default Min Replicas in case HPA mentioned, if not mentioned go with nodeSpec.Replicas mentioned
* `tiny-cluster-hpa.yaml` - Doc file for testing the feature.


deploy/crds/druid.apache.org_druids_crd.yaml
```diff
@@ -39,4 +39,4 @@ spec:
   versions:
   - name: v1alpha1
     served: true
-    storage: true
+    storage: true
```
himanshug: nit: not needed

examples/tiny-cluster.yaml
```diff
@@ -36,29 +36,24 @@ spec:
...
     druid.zk.service.compress=false
-
     # Metadata Store
     druid.metadata.storage.type=derby
     druid.metadata.storage.type=derby
     druid.metadata.storage.connector.connectURI=jdbc:derby://localhost:1527/var/druid/metadata.db;create=true
     druid.metadata.storage.connector.host=localhost
     druid.metadata.storage.connector.port=1527
     druid.metadata.storage.connector.createTables=true
-
```
himanshug: nit: no need to remove the newlines separating logical sets of properties

pkg/apis/druid/v1alpha1/druid_types.go
```diff
@@ -185,6 +187,9 @@ type DruidNodeSpec struct {
 	// Optional
 	ReadinessProbe *v1.Probe `json:"readinessProbe,omitempty"`
 
+	// Optional
+	AutoScale *autoscalev2beta1.HorizontalPodAutoscalerSpec `json:"autoscale,omitempty"`
```
himanshug:
```suggestion
	AutoScaler *autoscalev2beta1.HorizontalPodAutoscalerSpec `json:"autoscaler,omitempty"`
```

pkg/controller/druid/handler.go
```diff
@@ -59,6 +61,7 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 	serviceNames := make(map[string]bool)
 	configMapNames := make(map[string]bool)
 	podDisruptionBudgetNames := make(map[string]bool)
+	hpaName := make(map[string]bool)
```
himanshug:
```suggestion
	hpaNames := make(map[string]bool)
```

pkg/controller/druid/handler.go
```diff
@@ -151,6 +154,18 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
...
+		if nodeSpec.AutoScale != nil {
+			if err := sdkCreateOrUpdateAsNeeded(sdk,
+				func() (object, error) {
+					return makeHorizontalPodAutoscaler(&nodeSpec, m, ls, nodeSpecUniqueStr)
+				},
+				func() object { return makeHorizontalPodAutoscalerEmptyObj() },
+				nil, m, hpaName); err != nil {
+				return err
+			}
+		}
```
himanshug: can you also add section like `updatedStatus.StatefulSets = deleteUnusedResources(sdk, m, statefulSetNames, ls,...)`  for autoscaler resource so that it gets GC'd when not needed and also correctly reflected in status, See later for similar snippets for other resources we have?

AdheipSingh:
@himanshug i'll surely add the PR according to the template. TO answer your question, regarding the operator reconciling the replicas as mentioned originally in the spec... YES. 
I had tested this PR on existing cluster, the HPA Specs get created as desired but i din't test the how it behaves accordingly with respect to the operator, since i hardly had any load present.  A bit foolish on my part. Let me get back to you with stimulating ingestion and test the feature out. For the workaround, we shall need to add a helper function such as this 

```
func GetHPAReplicaCountOrDefault(client client.Client, name types.NamespacedName, defaultReplicaCount int32) int32 {
	var hpa autoscalev2beta1.HorizontalPodAutoscaler
	err := client.Get(context.Background(), name, &hpa)
	if err != nil {
		return defaultReplicaCount
	}

	if hpa.Spec.MinReplicas != nil && hpa.Status.DesiredReplicas < *hpa.Spec.MinReplicas {
		return *hpa.Spec.MinReplicas
	}

	return hpa.Status.DesiredReplicas
}
```

And use this function in place of replicas such as this 
```
Spec: appsv1.StatefulSetSpec{
			ServiceName:         serviceName,
			Replicas:            GetHPAReplicaCountOrDefault(Client, types.NamespacedName{
				Name:      hpaName(t),
				Namespace: r.Config.Namespace,
			}
```

NOTE: This is just pseudo code i came up with.

PR #15, opened 2020-02-16
Hpa Support for Spec

AdheipSingh:
Fixes #15 HPA support in CR Spec

### Description

Goal is to make the operator support Horizontal Pod Autoscaling for scaling druid pods.Add support in the druid node spec, so any user of this operator can have the flexibility of specifying 

Add support in the druid node spec, so any user of this operator can have the flexibility of specifying HPA on which ever node, which feels as necessity.

The node replicas should first check if hpa is deployed and get min replica count from the HPA spec. This way in case HPA scales the pods, the operator does not reconcile to its  state. 

As per my testing this what i have observed, When HPA is defined in the Spec, the pods are launched with the value minReplica as defined, and scale with load. The operator does not reconcile the state back. Which is what is right. When HPA not defined, nodeSpec.Replicas works as it was.

<hr>

This PR has:
- [ x] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
- [ x] added documentation for new or modified features or behaviours.

<hr>

 * `GetHPAReplicaCountOrDefault` - Func Added for getting the replica count in case HPA is used.

 * `Replicas: IntPointer(GetHPAReplicaCountOrDefault(sdk, types.NamespacedName{
				Name:      serviceName,
				Namespace: m.Namespace,
			}, nodeSpec.Replicas,` - Replicas to check default Min Replicas in case HPA mentioned, if not mentioned go with nodeSpec.Replicas mentioned
* `tiny-cluster-hpa.yaml` - Doc file for testing the feature.


himanshug:
thanks, I will check this again this week.

pkg/controller/druid/handler.go
```diff
@@ -151,6 +154,18 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 				return err
 			}
 		}
+
+		// Create HPA Spec
+		if nodeSpec.AutoScaler != nil {
```
himanshug: can we create this before  creating the StatefulSet to potentially StatefulSet to know about  it as soon as that gets created ?

pkg/controller/druid/handler.go
```diff
@@ -879,6 +926,26 @@ func namespacedName(name, namespace string) *types.NamespacedName {
 	return &types.NamespacedName{Name: name, Namespace: namespace}
 }
 
+// IntPointer returns a pointer to int32.
+func IntPointer(i int32) *int32 {
+	return &i
+}
+
+// GetHPAReplicaCountOrDefault returns the count of pods in case hpa is mentioned.
+func GetHPAReplicaCountOrDefault(client client.Client, name types.NamespacedName, defaultReplicaCount int32) int32 {
```
himanshug: So, I read a bit more about horizontal pod autoscaler and I think this shouldn't be needed. From what  I  understood, replica count set  on StatefulSet  doesn't matter when an autoscaler  is attached  to that StatefulSet.

Did you really observe operator  continuously updating the  StatefulSet replicas when autoscaler chose a different number of  replicas ?
or what goes wrong   if we  don't  do this?

examples/tiny-cluster-hpa.yaml
```diff
@@ -0,0 +1,193 @@
...
+        - name: data-volume
+          emptyDir: {}
+      resources:
+        requests:
+          memory: "2G"
+          cpu: "2"
+        limits:
+          memory: "2G"
+          cpu: "2"
+      autoscaler:
```
himanshug: I don't really think  that autoscaling historical nodes would be a good idea in production due to  coordinator constantly getting confused with segment management, having this in the example might make users believe that it is  a good practice  .

pkg/apis/druid/v1alpha1/druid_types.go
```diff
@@ -185,6 +187,9 @@ type DruidNodeSpec struct {
 	// Optional
 	ReadinessProbe *v1.Probe `json:"readinessProbe,omitempty"`
 
+	// Optional
+	AutoScaler *autoscalev2beta1.HorizontalPodAutoscalerSpec `json:"autoscaler,omitempty"`
```
himanshug: I noticed there is a thing  called "Vertical Pod Autoscaler" (in early development) as well, so it might  make more sense to call change this to...
```suggestion
	HPAutoScaler *autoscalev2beta1.HorizontalPodAutoscalerSpec `json:"hpAutoscaler,omitempty"`
```

so that, in future, we could add a `VPAutoScaler` too.

pkg/controller/druid/handler.go
AdheipSingh: @himanshug 
-> As mentioned in PR, with HPA deployed, hpa controller is responsible for maintaining replica count. The operator does not reconcile it back. 
-> Why we need this function is, if i do any change in CR lets say adjust limits and apply it, the replica count mentioned in the CR spec get deployed.....hpa again brings it back to the count mentioned in hpa spec. 
-> Their is shuffle pods, and recreation takes place. Plus CR should be the only spec file need to be maintained for the whole cluster. And the count should reflect both in CR as well as HPA at all times exactly the same.
-> With removing this function, this can happen. CR is showing 2 count for middlemanager ( as initially we deployed with 2 replica) and with hpa enabled replica is showing 5 ( as minreplica count is 5 ).
Here is why i feel that this function is important.

pkg/apis/druid/v1alpha1/druid_types.go
AdheipSingh: valid agreed

examples/tiny-cluster-hpa.yaml
AdheipSingh: valid agreed

AdheipSingh: @himanshug will VPA scaler work for historicals ? For middle managers HPA seems fine. Can you suggest which is suitable for historicals ?

pkg/controller/druid/handler.go
AdheipSingh: hpa needs to know about the statefulset, the ``` scaleTargetRef``` specifies which target to look to.  So what i feel ideally is a target gets created ( in our case its statefullset ) and then hpa gets attached to that target.

examples/tiny-cluster-hpa.yaml
AdheipSingh: Done Changes

pkg/controller/druid/handler.go
AdheipSingh: @himanshug a test is failing due this function. Not sure if the issue is in function, it's running fine in cluster. Did you find any specific issue in the function ?

himanshug: I think build is failing because stateful construction now requires a get to k8s  server to 
 succeed which doesn't exist when unit tests run.

I think main question is, is there any significance to `nodeSpec.replicas` attribute , wouldn't k8s just ignore that on the StatefulSet when HPA exists for that StatefulSet ? And, in case, we need to set that to min replica value as in HPA spec, then why goto k8s api server to get it ..rather read it from the hpa spec we already have in the cr ?

AdheipSingh: @himanshug Yes when ```minRelicas``` declared in the spec, k8s would take that as source of truth for the replica count. As per my practice, i don't mention the replica count in the spec when using HPA as its more confusing to have different counts. 
At the same time your point seem LGMT we can take the count from the spec itself. But i still feel as per practice no need to mention replicas when using HPA. 
Your call either works.

himanshug: we definitely shouldn't contact the k8s server to get hpa when we already have it.

on whether to try/adjust same on StatefulSet , I think it is ok to not change anything for StatefulSet, users know that they have configured hpa and that replica count on StatefulSet is ignored by k8s in this case. It would be same experience if user deployed Druid cluster without using this operator.

AdheipSingh: Agreed, i'll remove the client context part and get the value directly from HPA spec.

AdheipSingh: Done changes, now not using client context for getting the default count. Getting directly from ```nodeSpec.Replicas```. I have tested this on my cluster, works for me.

pkg/controller/druid/handler.go
```diff
@@ -879,6 +926,28 @@ func namespacedName(name, namespace string) *types.NamespacedName {
...
 }
 
+// IntPointer returns a pointer to int32.
+func IntPointer(i int32) *int32 {
+	return &i
+}
+
+// GetHPAReplicaCountOrDefault returns the count of pods in case hpa is mentioned.
+func GetHPAReplicaCountOrDefault(name types.NamespacedName, defaultReplicaCount int32) int32 {
+	var hpa autoscalev2beta1.HorizontalPodAutoscaler
```
himanshug: this doesn't appear to be initialized anywhere  according to what we have in CR ?

pkg/controller/druid/handler.go
```diff
@@ -879,6 +926,28 @@ func namespacedName(name, namespace string) *types.NamespacedName {
...
+	var nodeSpec v1alpha1.DruidNodeSpec
+
+	err := &nodeSpec.Replicas
+	if err != nil {
+		return defaultReplicaCount
```
himanshug: other validator PR  *requires* `nodeSpec.Replicas` so , this case  will always be true once that  PR  is merged.

pkg/controller/druid/handler.go
AdheipSingh: dint get you. its just initialized in this function

himanshug: this line initializes it to empty HorizontalPodAutoscaler struct , and then it is directly used at line 944 , we don't seem to be reading the actual value that user might have put in the CR or  are you just trying to set to default min value of empty `HorizontalPodAutoscaler` for which `hpa.Spec.MinReplicas` is always nil and `hpa.Status.DesiredReplicas` is always 0.

AdheipSingh: Yes, replica will show different than minreplicas i guess thats what we agreed on. Plus hpa.Sepc.MinReplicas will show what is mentioned not zero. Here's a snippet 
```
"apiVersion\":\"apps/v1\"},\"minReplicas\":1,\"maxReplicas\":10,\"metrics\":[{\"type\":\"Resource\",\"resource\":{\"name\":\"cpu\",\"targetAverageUtilization\":60}}]},\"status\":{\"currentReplicas\":0,\"desiredReplicas\":0,\"currentMetrics\":null,\"conditions\":null}}","name":"tiny-cluster","namespace":"default"}
```
Plus to clarify this way the whole pod creation takes place, the stateful set is created with replica 1 ( assume we mention replica 1) , replica one get created, HPA get deployed ( assume we mention 3 as minReplica ), after seeing the desired to replica to 3. Hpa takes replica to count 3. 
Earlier with client, in case hpa mentioned it would directly take count to 3.

AdheipSingh: Seems this function is really making no sense. This would only make sense in case ```replicas *int32```, since replica is required and is absolute no point having this. I guess its overdo.  I get it this let me muke this. Lets keep it straightforward. @himanshug

himanshug: Yeah, I would prefer to just  let the StatefulSet be as it was before and not try to adjust replica count on statefulSet at all.

pkg/controller/druid/handler.go
```diff
@@ -879,6 +922,11 @@ func namespacedName(name, namespace string) *types.NamespacedName {
 	return &types.NamespacedName{Name: name, Namespace: namespace}
 }
 
+// IntPointer returns a pointer to int32.
+func IntPointer(i int32) *int32 {
+	return &i
+}
+
```
himanshug:
```suggestion
```

AdheipSingh: sorry i am having a bad day at work.

AdheipSingh: How come now CI is failing ? Ran go fmt local it works.

himanshug: gofmt did not like the extra empy line introduced at line#925 after  latest commit.

PR #28, opened 2020-03-20
ServiceAccount and TerminationGracePeriod

### LINKED ISSUE #10: TerminationGracePeriod Support
Currently termination grace period is set to default ie 30 sec. Can we add support in the spec so that we can modify the grace period.

### LINKED ISSUE #24: Deploy druid cluster in Service Account and support ImagePullSecrets
I propose to deploy the druid cluster in a service account so that adequate RBAC's can be applied in a prod environment. 
Currently operator cannot pull custom built images private registeries which require authentication, add support for ImagePullSecrets option in sts.
Currently only the operator runs in a SA. Not sure  adding SA in druid cluster spec should it be a Required or Optional Parameter ? 
@himanshug could you suggest on service account

AdheipSingh:
Fixes #10 
Fixes #24 

### Description

TerminationGrace Period is limited to 30sec as default. Some processes such as middle managers, historicals needs more graceful time during an update.
ServiceAccount, deploys the cluster in a specified SA, helps in giving access, applying RBACs to druid cluster across a cluster

- This func returns the TGP if defined else keeps is too 30. When directly using TGP without the func k8s keeps the sts TGP to 0. 
```
func getTerminationGracePeriod(nodeSpec *v1alpha1.DruidNodeSpec) 
```
<hr>

This PR has:

<hr>

 * ```handler.go```
 * ```druid_types.go```

### DIFF
--- a/pkg/apis/druid/v1alpha1/druid_types.go
+++ b/pkg/apis/druid/v1alpha1/druid_types.go
@@ -2,6 +2,7 @@ package v1alpha1
 
 import (
 	"encoding/json"
+
 	autoscalev2beta1 "k8s.io/api/autoscaling/v2beta1"
 
 	appsv1 "k8s.io/api/apps/v1"
@@ -9,6 +10,10 @@ import (
 	"k8s.io/api/policy/v1beta1"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	DefaultTerminationGracePeriod = 30
 )
 
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
@@ -55,6 +60,9 @@ type DruidClusterSpec struct {
 	// Required: Druid Docker Image
 	Image string `json:"image"`
 
+	// Optional: ServiceAccount for the druid cluster
+	ServiceAccount string `json:"serviceAccount"`
+
 	// Optional: environment variables for druid containers
 	Env []v1.EnvVar `json:"env,omitempty"`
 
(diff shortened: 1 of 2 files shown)


pkg/controller/druid/handler.go
```diff
@@ -675,6 +675,15 @@ func updateDefaultPortInProbe(probe *v1.Probe, defaultPort int32) *v1.Probe {
 	return probe
 }
 
+func getTerminationGracePeriod(nodeSpec *v1alpha1.DruidNodeSpec) *int64 {
+
+	if nodeSpec.TerminationGracePeriodSeconds == 0 {
+		nodeSpec.TerminationGracePeriodSeconds = 30
```
himanshug: can we leave it be default value that k8s cluster would have instead of hardcoding 30 secs in the operator

AdheipSingh: I had kept to absolute value, changed type to pointer.

pkg/apis/druid/v1alpha1/druid_types.go
```diff
@@ -157,6 +161,9 @@ type DruidNodeSpec struct {
 	// Optional: affinity to be used to for enabling node, pod affinity and anti-affinity
 	Affinity *v1.Affinity `json:"affinity,omitempty"`
 
+	// Optional: terminationGracePeriod defaults to 30 sec.
```
himanshug:
```suggestion
	// Optional: terminationGracePeriod
```

AdheipSingh: Done

PR #33, opened 2020-03-28
Helm Chart for deploying operator

AdheipSingh:
### Description

- Provide helm chart to deploy to operator.
- Currently need to manually deploy all the manifest, both should be supported

- Helm chart shall deploy service account, rbac and operator deployment.
- Helm chart shall support further configuration parameter such as selector and affinities
- The default image shall point to the latest release which 0.0.1 in the current phase.
<hr>

This PR has:
- [ x] added documentation for new or modified features or behaviors.

<hr>

 * `charts` - contains the helm chart and documentation
 * `README` - added support for both helm based installation


himanshug:
I  am  not a helm user and this might be a  naive question. But,  I fail to understand  what  is the value we get out of helm in this case.

If the  problem  is that manifests need 4 commands i.e.
```
druid-operator$ kubectl create -f deploy/service_account.yaml
# Setup RBAC
druid-operator$ kubectl create -f deploy/role.yaml
druid-operator$ kubectl create -f deploy/role_binding.yaml
# Setup the CRD
druid-operator$ kubectl create -f deploy/crds/druid.apache.org_druids_crd.yaml
```

then 4 manifests above can be put inside a single yaml  file allowing single `kubectl create  -f ..`  creating all  4 of those.

AdheipSingh:
Major Operators support manual installation using manifests, helm, kustomize and ansible installation. The value added can be seen in the values.yaml, where we have more flexibility of deploying such as affinites, resource etc. Its not about just deploying but maintaing version using Charts.yaml, in large orgranizations such charts definitely add value in maintaince and ugrades of operators. :)

himanshug:
sorry, forgot about this.

that said, I see that there are many different tools to install the operator itself in addition plain k8s yamls. I looked at few other operators, and it seems like  https://github.com/helm/charts/tree/master/stable is the right home for such community maintained helm charts and is used by various operators. more importantly, that repository likely has right  set of people to review the charts and offer suggestions.

AdheipSingh:
https://github.com/helm/charts#deprecation-timeline

AdheipSingh:
> sorry, forgot about this.
> 
> that said, I see that there are many different tools to install the operator itself in addition plain k8s yamls. I looked at few other operators, and it seems like https://github.com/helm/charts/tree/master/stable is the right home for such community maintained helm charts and is used by various operators. more importantly, that repository likely has right set of people to review the charts and offer suggestions.

https://github.com/helm/charts#deprecation-timeline

AdheipSingh:
@himanshug post your webinar, i did see some attraction towards helm, let me know if you would like to bring this up.

Plus we can also wrap up the CR in template in helm charts and leverage manual rollbacks of helm too.

himanshug:
yeah, if you can test and confirm that it is working , then we can merge it.

charts/templates/rbac.yaml
```diff
@@ -0,0 +1,110 @@
...
+apiVersion: rbac.authorization.k8s.io/{{ required "A valid .Values.rbac.apiVersion entry required!" .Values.rbac.apiVersion }}
+kind: Role
+metadata:
+  name: {{ template "druid-operator.fullname" . }}
+  labels:
+    app: "{{ template "druid-operator.fullname" . }}"
+    chart: "{{ include "druid-operator.chart" . }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+rules:
```
davejachimiak, line 11: I think this needs a rule for `ingresses.extensions` reflected by [this change](https://github.com/druid-io/druid-operator/commit/0372b5d17b82d2a823317f5b280c32681738ed17). Otherwise, attempts by the operator to create  workload objects will fail.

PR #42, opened 2020-05-02
LifeCycle object

### LINKED ISSUE #37: Protect Middle Manager From Deletion when a running task is associated with MM
Scenario: 
MM are running, and incase of dynamic scaling their are time when HPA scales down the MM, even when MM are running a task. Even regardless of HPA their should be a mechanism which protects the MM Pod from getting deleted, when a task is associated with that MM at all Times.

Not exactly sure on how to solve this, but maybe having finalizers can be useful. Finalizers can make sure a pre-deletion hooks runs and make sure the pod remains in terminating phase till the time task is not complete ( not sure how feasible this options is as i see mostly finalizers are associated with CRD's, how can i associate it with statefulset). 

Have pre-Stop hook and extend the TerminationGracePeriod ( which is supported by operator ) to make sure the tasks are running. 

Some questions arise:
How do we me make a mapping  which between a MM pod and task being associated with it. The endpoint ```/druid/v1/indexer/runningTasks``` has the following json output
```
[{"id":"ds_2020-04-09T17:41:09.220Z","groupId":"ds",09T17:41:09.220Z","type":"compact","createdTime":"2020-04-09T18:19:05.795Z","queueInsertionTime":"2020-04-09T18:19:05.795Z","statusCode":"RUNNING","status":"RUNNING","runnerStatusCode":"RUNNING","duration":null,"location":{"host":"10.28.6.38","port":8100,"tlsPort":-1},"dataSource":"ds","errorMsg":null}]
```
Here the ```"host":"10.28.6.38"``` maps to a kubernetes ep ( ```kubectl get ep```). Now i need to make sure anyhow the pod associated with this host does not scale down or gets deleted at all times.

Any suggestions, ideas would be helpful.
@vDMG

AdheipSingh:
Fixes #37

### Description

- Prestop hook can run a script to perform certain tasks. Example a preStop bash script can make sure not delete a middlemanager if its a running a task. Here's a sample, this while loop only exits if no other tasks apart from root process in running. 
- To run hooks, bake them as a command in the image and execute the command.
```
#!/bin/bash
SERVICE="java"
while :
do
RESULT=`pgrep -x $SERVICE | sed '1d'`
if [[ "$RESULT" != "" ]];then
    echo "Peon Running"
else
    echo "Peon Not Running"
    break
fi
sleep 5
done
```

<hr>

This PR has:

<hr>

 * `handler.go`
 * `types.go`
 * `deepcopy.go`


himanshug:
can we  add support for whole lifecycle object  instead ?

AdheipSingh:
sure :)

ChethanUK:
> * Prestop hook can run a script to perform certain tasks. Example a preStop bash script can make sure not delete a middlemanager if its a running a task. Here's a sample, this while loop only exits if no other tasks apart from root process in running.

Cool

pkg/apis/druid/v1alpha1/druid_types.go
```diff
@@ -200,6 +200,9 @@ type DruidNodeSpec struct {
 	// Optional
 	ReadinessProbe *v1.Probe `json:"readinessProbe,omitempty"`
 
+	// Optional
+	Lifecycle *v1.Lifecycle `json:"preStop,omitempty"`
```
himanshug, line 204: I  think you probably forgot to rename `preStop` to `lifeCycle` ?

AdheipSingh: oh :(

AdheipSingh: ill correct this

pkg/apis/druid/v1alpha1/druid_types.go
```diff
@@ -206,6 +206,9 @@ type DruidNodeSpec struct {
 	// Optional
 	ReadinessProbe *v1.Probe `json:"readinessProbe,omitempty"`
 
+	// Optional
+	Lifecycle *v1.Lifecycle `json:"lifecycle,omitempty"`
```
himanshug, line 210: where is this used in handler now ?

himanshug:
pls resolve conflicts.

AdheipSingh:
hmm fixing CI....

himanshug:
@AdheipSingh   This PR  is merged, but  please update the Description which is outdated .

Issue #35, opened 2020-03-31
Allow possibilities to add env var to erase some config value


vDMG:
## Context
Today, all our password are in clear in our configuration, which is not a good securiy practice.
**example**
```
druid.metadata.storage.connector.password=toto
druid.zk.service.pwd
```

## Proposal

We could allow user to specify somme environment variable to erase some values
**example**
```
env:
- name: DRUID_METADATA_STORAGE_CONNECTOR_PASSWORD
   valueFrom:
     secretKeyRef:
       name: druid-connector
       key: password
```

We could imagine to replace after those values in the configuration.
We need to define stricts norms for those environment variables ("\_" replace ".",all upper case and starting with "DRUID_".

----
 
What are your thoughts about this idea ? @himanshug @AdheipSingh

AdheipSingh:
Hey @vDMG 
Here is a workaround
```
druid.metadata.storage.connector.password={"type":"environment","variable":"CONNECTOR_KEY"}
```
then use this as an env
```
env:
  - name: CONNECTOR_KEY
    valueFrom:
      secretKeyRef:
        key: db-password
        name: druid-secrets
```

vDMG:
Actually the official druid docker image handles it perfectly using env vars like `druid_metadata_storage_connector_password` replaced on the service (overlord, historical,...) configuration with `druid.metada.storage.connector.password`. 👍

himanshug:
I think https://github.com/druid-io/druid-operator/issues/35#issuecomment-606562407 is  the right way to handle password for  both db and zk.
support for adding password via  env variables exists in Druid to protect  those secrets and  works  in all deployments (even the ones without docker and/or k8s)

PR #45, opened 2020-05-14
add templating feature to the operator

Myafq:
# Description

I have druid installation in dynamic environments with an external database, which requires credentials, and also I use an S3 connection. So it's important for me to use some secrets management. But currently druid operator doesn't provide any way to include secrets values to the druid object.

There multiple ways to approach this issue:
- We could add secrets directly to the Pods with druid. But in this case, we will require the support of those secrets from the image, which is additional work.
- We can dynamically access secrets during the render of configmaps. (which is implemented)
- In combination with the previous one, we also can create secrets instead of configmaps and mount them, if required. (but I decided that this feature can be the second iteration)

The main change of this patch - added field "templateValuesRefs" to the Druid spec. which can refer to confimap, secrets, or plain values. And then processing of the whole manifest through template render with those values. 
<hr>

This PR has:
- [✔️] been tested on a real K8S cluster to ensure the creation of a brand new Druid cluster works.
- [ ✔️] been tested for backward compatibility on a real K*S cluster by applying the changes introduced here on an existing Druid cluster. If there are any backward-incompatible changes then they have been noted in the PR description.
- [✔️ ] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
- [✔️ ] added documentation for new or modified features or behaviors.

<hr>

 * `pkg/apis/druid/v1alpha1/druid_types.go`
 * `pkg/controller/druid/handler.go`
 * `example`


AdheipSingh:
@Myafq https://github.com/druid-io/druid-operator/issues/35#issuecomment-608396033 in case you are looking to add secrets here a way around. :)

Myafq:
> @Myafq [#35 (comment)](https://github.com/druid-io/druid-operator/issues/35#issuecomment-608396033) in case you are looking to add secrets here a way around. :)

This feature is much more powerful than just that. You basically can template any string value in the manifest. I think it'll help with future features as well.

himanshug:
@Myafq while this is an interesting and very generic feature, but recommended way to achieve secrets management should be via Druid itself using the `PasswordProvider` interface as pointed in comments above.
Is there a particular real use case that is currently not possible without this feature and Druid itself is not suitable place for implementing that?

PR #46, opened 2020-05-21
Support for Ingress and Labels

### LINKED ISSUE #40: Add Ingress in Spec
Brokers and routers can be exposed using ingress, the operator should support ingress spec.

AdheipSingh:
Fixes #40 

### Description
- Ingress in nodeSpec, so that routers + brokers can be exposed as needed.
- PodLabels for clusterSpec, NodeSpec

<hr>

This PR has:
- [ ]x been tested for backward compatibility on a real K*S cluster by applying the changes introduced here on an existing Druid cluster. If there are any backward incompatible changes then they have been noted in the PR description.
- [ x] added documentation for new or modified features or behaviors.

<hr>

- handler.go
- types.go
-  tiny-cluster-ingress.yaml
- deepcopy.go


examples/tiny-cluster-ingress.yaml
```diff
@@ -0,0 +1,187 @@
+apiVersion: "druid.apache.org/v1alpha1"
```
himanshug, line 1: I  think we will end up with  a long list of confusing tiny-cluster specs  when adding one new file for each feature addition. I think adding documentation is probably better way to go at this point. so let us focus on  that instead of new spec . I think it  is mentioned in another ticket.

himanshug:
LGTM otherwise

examples/tiny-cluster-ingress.yaml
AdheipSingh: sure ill remove the spec

PR #57, opened 2020-06-19
[WIP]: Test operator against cluster

AdheipSingh:
- Work In Progress

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -16,7 +16,7 @@ spec:
       containers:
         - name: druid-operator
           # Replace this with the built image name
-          image: REPLACE_IMAGE
+          image: localhost:5000/druid-io:latest
           command:
           - druid-operator
           imagePullPolicy: Always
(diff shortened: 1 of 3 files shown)


AdheipSingh:
@himanshug 
- make test shall run unit tests if clear we proceed else fail.
- we create cluster using KIND ( kubernetes in Docker ), build docker image, and push to local registery ( push pulls are faster in local registery ).
- deploy operator, crd, cr and zookeeper.
- check all pods are in running state, if not exit 1.
- Add auto push to druid-io with tag when on branch master

the current implementation is just to get some look into how it coming up, will add some improvements. 
here the link to kind https://github.com/kubernetes-sigs/kind
Looking forward to suggestions and improvements in this approach.

deploy/operator.yaml
```diff
@@ -16,7 +16,7 @@ spec:
       containers:
         - name: druid-operator
           # Replace this with the built image name
-          image: REPLACE_IMAGE
+          image: localhost:5000/druid-io:latest
```
AdheipSingh, line 19: @AdheipSingh to be kept to REPLACE_IMAGE

AdheipSingh:
have smoke tests now

PR #70, opened 2020-08-13
PVC For deployments

AdheipSingh:
### Description

- PersistentVolumeClaims to be deployed by operator and mounted to deployments nodes. 
- To support MM which can be run as deployments + use full for heap dumps after pod killed for troubleshooting.

- PVC creation to take place before node deployed so that mount does not get stuck in.
- PVC name is kept to `nodeSpecUniqueStr` the operator manages volumeMounts and volumes. no need to manually add, just specify the MountPath during creation.

<hr>

This PR has:

<hr>

 * `types.go`
 * `handler.go`

### DIFF
--- a/pkg/apis/druid/v1alpha1/druid_types.go
+++ b/pkg/apis/druid/v1alpha1/druid_types.go
@@ -229,6 +229,11 @@ type DruidNodeSpec struct {
 	// Optional: Ingress Spec
 	Ingress *extensions.IngressSpec `json:"ingress,omitempty"`
 
+	// Optional: Persitance Volume Spec for Deployments
+	PersistentVolumeClaim *v1.PersistentVolumeClaimSpec `json:"persistentVolumeClaim,omitempty"`
+
+	// Optional: MounthPath to mount volume
+	PersistentVolumeClaimMounthPath string `json:"persistentVolumeClaimMountPath,omitempty"`
 	// Optional
 	Lifecycle *v1.Lifecycle `json:"lifecycle,omitempty"`
 
@@ -256,14 +261,15 @@ type DeepStorageSpec struct {
 }
 
 type DruidClusterStatus struct {
-	StatefulSets         []string `json:"statefulSets,omitempty"`
-	Deployments          []string `json:"deployments,omitempty"`
-	Services             []string `json:"services,omitempty"`
-	ConfigMaps           []string `json:"configMaps,omitempty"`
-	PodDisruptionBudgets []string `json:"podDisruptionBudgets,omitempty"`
-	Ingress              []string `json:"ingress,omitempty"`
-	HPAutoScalers        []string `json:"hpAutoscalers,omitempty"`
-	Pods                 []string `json:"pods,omitempty"`
+	StatefulSets          []string `json:"statefulSets,omitempty"`
+	Deployments           []string `json:"deployments,omitempty"`
+	Services              []string `json:"services,omitempty"`
+	ConfigMaps            []string `json:"configMaps,omitempty"`
+	PodDisruptionBudgets  []string `json:"podDisruptionBudgets,omitempty"`
+	Ingress               []string `json:"ingress,omitempty"`
+	PersistentVolumeClaim []string `json:"persistentVolumeClaim,omitempty"`
+	HPAutoScalers         []string `json:"hpAutoscalers,omitempty"`
+	Pods                  []string `json:"pods,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/controller/druid/handler.go
+++ b/pkg/controller/druid/handler.go
@@ -68,6 +68,7 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 	podDisruptionBudgetNames := make(map[string]bool)
 	hpaNames := make(map[string]bool)
 	ingressNames := make(map[string]bool)
+	pvcNames := make(map[string]bool)
 
 	ls := makeLabelsForDruid(m.Name)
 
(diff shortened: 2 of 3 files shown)


AdheipSingh:
+ cc @nishantmonu51

AdheipSingh:
@himanshug any update ? :)

pkg/apis/druid/v1alpha1/druid_types.go
```diff
@@ -229,6 +229,11 @@ type DruidNodeSpec struct {
 	// Optional: Ingress Spec
 	Ingress *extensions.IngressSpec `json:"ingress,omitempty"`
 
+	// Optional: Persitance Volume Spec for Deployments
+	PersistentVolumeClaim *v1.PersistentVolumeClaimSpec `json:"persistentVolumeClaim,omitempty"`
+
+	// Optional: MounthPath to mount volume
+	PersistentVolumeClaimMounthPath string `json:"persistentVolumeClaimMountPath,omitempty"`
```
himanshug, line 236: I am slightly concerned over existence of these two and volumeXXX keys that we already had for StatefulSet. I haven't explored in detail but it would be good to somehow, at spec level, have only one set of keys regarding the persistent volumes and that be used for both StatefulSet and Deployment.

AdheipSingh: i had the same query :) , since both have  ```type persistentVolumeClaim```,
 The reason i went was just to not have confusion and follow as  per k8s convention deployments have pvc mounts and sts have volumeclaimtemplates. In case you want to have a common name, it should not conflict with k8s convention. 
open to suggestions :)

pkg/controller/druid/handler.go
```diff
@@ -115,6 +116,15 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 			return err
 		}
 
+		if nodeSpec.PersistentVolumeClaim != nil && nodeSpec.Kind == "Deployment" {
```
AdheipSingh, line 119: I wonder why are we restricting PVC creation to Deployment only ? I came across a case where i want to use sts with multiple pvc mounts for heap dumps. with a shared storage class. Ideally operator should support both.

himanshug:
sorry, this is probably a step back, but I am trying to understand the behavior when persistent volume is used together with Deployment which is thought of something for "stateless" things, here are a few gotchas when persistent volume is used with Deployment 

1. Unlike StatefulSet, mounted volume across different replica pods of a Deployment would get the same shared volume and not one dedicated for the specific pod . It is ok for heapdump use case (as long as user ensures that heapdump from different peon processes even in different pod would have a different name) but still a catch. Also, it might be very unintuitive/error-prone if user inadvertently uses persistent volume with Deployment for some other use case.

2. you can not scale Deployment when using persistent volume,
Ref: https://akomljen.com/kubernetes-persistent-volumes-with-deployment-and-statefulset/
Is (2) not a hard limitation because user would definitely be scaling MMs up/down time to time ?

AdheipSingh:
For heapdump path conflict, use env $HOSTNAME inside the pod, so each dump is unique, like this
```-XX:HeapDumpPath=/opt/druid/heapdumps/$HOSTNAME``` 

I guess  pvc adds complexity as well as flexibility :)

Plus using shared storage class for dumps like NFS works well, main catch is your storage class provider should support ```AccessMode: ReadWriteMany``` so that pvc can be mounted even if we scale to multiple pods.

Regarding MM i guess you can answer best ( i somehow believe in the the best way to run MM is by just using SIMPLE POD AND PVC , no deployments no sts, each pod/pvc is unique having their own svc, atleast i can turn on and off MM easily )

Assume we use deployments with MM so in case we scale MM to 2 and both pods  have a shared pvc...... will their be a conflict ?

**Comment by himanshug** [comment]
> Assume we use deployments with MM so in case we scale MM to 2 and both pods have a shared pvc...... will their be a conflict or it works fine ?

there is conflict when PVC is used with Deployment resource since all pods get same space unlike StatefulSet .

Currently, for which node type are you using `Deployment` resource and is it possible to just use StatefulSet there ?

**Comment by AdheipSingh** [comment]
I am using deployment for all expect MM and historicals ( hot and cold ). Might shift MM to deployments

I did check the MM issue with other druid members, running 2 MM should be fine even if they share the pvc. So might even run MM with deployments. So sharing pvc space shouldnt be an issue for deployments, the link you mentioned above 

Assuming a user will always use Historicals with statefulset
Mainly the purpose of this PR is ( though their is some more clean up to be done )

1. User can add pvc to deployments - for heapdumps, or any other case, you can scale to N pods with shared pvc. No conflict should be their
2. User can add pvc to sts apart from default volumeclaimtemplate

So ideally operator provisions pvc, user can mount wherever however they want.

**Comment by AdheipSingh** [comment]
@himanshug let me know if you think this is feasible, though i have been using this for sometime, might be very specific to usecase, please feel free to close if its not feasible from community approach

**Comment by himanshug** [comment]
one workaround you/users have is that you can create your PVC outside of the operator (probably using the helm chart that you use to deploy the operator) and then refer to that PVC using existing `volumeMounts` and `volumes` keys in nodeSpec where you used the `deployment`.

above should be good enough and hopefully that is what you are doing.

-----------

however, if above is really useful and we want to automate PVC creation in druid-operator itself............ next part is about how to do that in a slightly more general way...

we already have (both at top level and at individual nodeSpec level) ...

```
	VolumeClaimTemplates []v1.PersistentVolumeClaim `json:"volumeClaimTemplates,omitempty"`
	VolumeMounts         []v1.VolumeMount           `json:"volumeMounts,omitempty"`
	Volumes              []v1.Volume                `json:"volumes,omitempty"`
```

then you are adding following at the top level...

```
	// Optional: Persitance Volume Spec for Deployments
	PersistentVolumeClaim *v1.PersistentVolumeClaimSpec `json:"persistentVolumeClaim,omitempty"`

	// Optional: MounthPath to mount volume
	PersistentVolumeClaimMounthPath string `json:"persistentVolumeClaimMountPath,omitempty"`
```

so the proposal above adds ability to tell operator to create ONE pvc and mount that only in `deployment` resources .

Instead we could add (at top level)

```
persistentVolumeClaims  []v1.PersistentVolumeClaim `json:"persistentVolumeClaims,omitempty"` 
```

and operator would just submit those to k8s blindly, and user can then use those PVCs by adding them to the existing  `volumes` and `volumneMounts` part of current spec. and, we wouldn't do any "if-deployment" kind of checks ... those PVCs could be used wherever user wants to use them.

AdheipSingh:
i see, got your point. Lets keep it straightforward then ill add separate PR for this. 
Thanks for looking into this :)

PR #80, opened 2020-09-16
deletepod when sts hangups

### LINKED ISSUE #79: Pod Not Getting Re-Loaded After CR Updated
- The operator maintains a hash which is associated at the metadata annotation level and at configMap in  env vars of pods.
- Mainly to reload the configs if any change happens in the any of the object  reconciled by the operator ( any object using runtime.object interface )

ISSUE

- I applied a wrong config to my Druid CR which resulted in cold historical node running as sts to fail, which led to crashing state. The operator rightly halt the updates since rolling Deploy was true.
- After i applied the right configuration to Druid CR their was no trigger of a new sts or pod by the operator. I had to manually delete the sts for the new configs to be applied. 

- I am curious why dint the operator re-load the config, since on debug the prev hash and new hash were different. 
- I believe the annotations for hash should be maintained at the pod level ( pod template metadata ), i havent tested out this approach, but i see that helm is also hashing configurations at the pod level and not sts/deploy metadata.

- Not sure if this is a desired behavior or something wrong in the operator. 

NOTE: This behavior i only see for sts when on crashing state, after new configs applied the pod still remains in crashing state and does not restart/reload the new cm updates

@himanshug

AdheipSingh:
Fixes #79

### Description

- checkstatus() will check the status of the pod in case its in crashstate, it will delete it, this way we are able to reload the new config as discussed in the issue #79 
- Have kept the restart count > 1 otherwise the function will not let any pod come up and keep on deleting that pod.
- Few questions/issues open to discuss 

1. The function is being called after sts updates, i felt this is the right place to call function since after the stsFullyDeployed() can make sure if the update is fully success or not.  
2. The issue here is in case the error is not pertaining to sts controller hanging up, this behavior might cause some issue, since it keeps on deleting the sts till the time we fix it. 
3. From a community perspective ,Shall we keep this optional or default ?

This has been tested in #79 scenario and gave +ve outcome :smile: 
<hr>

This PR has:

<hr>

 * `handler.go`


AdheipSingh:
CI is failing regardless of my PR, i guess with the add of #74 is causing, will debug and fix this later :|

pkg/controller/druid/handler.go
```diff
@@ -174,7 +174,11 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 				if !done {
 					return err
 				}
+
 			}
+
+			checkCrashStatus(sdk, m)
+
```
AdheipSingh, line 181: This should ideally be true only for sts with podmanagement policy orderready, since its only for orderready.

AdheipSingh: @himanshug in case you get time, do review this one :)

pkg/controller/druid/handler.go
```diff
@@ -328,6 +332,37 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 	return nil
 }
 
+func checkCrashStatus(sdk client.Client, m *v1alpha1.Druid) {
+	podList := podList()
+	if err := sdk.List(context.TODO(), podList); err != nil {
```
himanshug, line 337: this would list all pods in namespace where operator is running? instead we would want to list all pods in the user spec's namespace in a given statefulSet .. no?

pkg/controller/druid/handler.go
```diff
@@ -328,6 +332,37 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
...
+	}
+
+	pod := make([]*v1.Pod, 0)
+	for i := range podList.Items {
+		pod = append(pod, &podList.Items[i])
+	}
+
+	for _, p := range pod {
+		if p.Status.ContainerStatuses[0].RestartCount > 1 {
+			for _, condition := range p.Status.Conditions {
```
himanshug, line 350: I am not entirely sure about how this works but, I am assuming it has been tested well and known to work. :)

pkg/controller/druid/handler.go
```diff
@@ -174,7 +174,11 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 				if !done {
 					return err
 				}
+
 			}
+
+			checkCrashStatus(sdk, m)
```
himanshug, line 180: can we hide this feature behind a flag and default to true?

also, for rolling deploy enabled case, this needs to happen before line 173.

AdheipSingh: sure, i had that doubt if this should be flagged, will do that.

pkg/controller/druid/handler.go
```diff
@@ -328,6 +332,37 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
...
+	for i := range podList.Items {
+		pod = append(pod, &podList.Items[i])
+	}
+
+	for _, p := range pod {
+		if p.Status.ContainerStatuses[0].RestartCount > 1 {
+			for _, condition := range p.Status.Conditions {
+				if condition.Type == "Ready" {
+					if p.Status.Phase != "Running" || condition.Status == "False" {
+						err := sdkDelete(context.TODO(), sdk, p)
```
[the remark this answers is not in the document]
AdheipSingh: ok agreed, will do

pkg/controller/druid/handler.go
AdheipSingh: took my almost a day to figure out what condition matches exactly crashloopback :)

pkg/controller/druid/handler.go
AdheipSingh: we will call this function inside the sts block plus will wrap this in an if condition such as 

if sts.Podmanagement == "orderReady" {
checkCrashStatys()
}

so should be good

himanshug: even then, I would say we should only ask to list pods that we are concerned about, not every pod ... we should be able to do that by using right namespace and label selectors.

AdheipSingh: sure agreed, can add namespace to []client.list but not sure what  label selector to use  unless a map[string]string exists specifically for filtering MM.

AdheipSingh:
@nishantmonu51 can you take a look at this too :smiley:

docs/upgrade.md
```diff
@@ -4,3 +4,5 @@
...
+- During upgrade if sts is set to ordered ready, the sts controller will not recover from crashloopback state. The issues is referenced [here](https://github.com/kubernetes/kubernetes/issues/67250), and here's a reference [doc](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#forced-rollback)
+- How operator solves this is using the ```forcedRollbackSts``` key, if this key is set to ```true``` the operator will delete the sts pod if its in crashloopback state. Example Scenario: During upgrade, user rolls out a faulty configuration causing the historical pod going in crashing state, user rolls out a valid configuration, the new configuration will not be applied unless user manual delete pods, so solve this scenario operator shall delete the pod automatically without user intervention. ```NOTE: User must be aware of this feature, their might be cases where crashloopback might be caused due probe failure, fault image etc, the operator shall keep on deleting on each re-concile loop```
```
nishantmonu51, line 8: s/their might be cases/there might be cases/g
Can you also mention the default value of this new flag in the docs ?

pkg/controller/druid/handler.go
```diff
@@ -169,12 +169,23 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
...
+				}
+
 				// Check StatefulSet rolling update status, if in-progress then stop here
 				done, err := isStsFullyDeployed(sdk, nodeSpecUniqueStr, m)
 				if !done {
 					return err
 				}
 			}
+			if m.Spec.ForcedRollbackSts == true {
+				if nodeSpec.PodManagementPolicy == "OrderedReady" {
```
nishantmonu51, line 185: minor nit: this can be translated to if A && B condition

pkg/controller/druid/handler.go
```diff
@@ -328,6 +339,44 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 	return nil
 }
 
+func checkCrashStatus(sdk client.Client, m *v1alpha1.Druid) {
+	podList := podList()
+	listOpts := []client.ListOption{
+		client.InNamespace(m.Namespace),
+	}
```
nishantmonu51, line 346: would this actually list all the pods in the namespace ?
There could be other pods in the namespace besides those managed by the operator itself. 
Is there a way to filter the pods which are managed by the druid-operator vs those not-managed by druid operator ?

pkg/controller/druid/handler.go
```diff
@@ -328,6 +339,44 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
...
+	pod := make([]*v1.Pod, 0)
+	for i := range podList.Items {
+		pod = append(pod, &podList.Items[i])
+	}
+
+	for _, p := range pod {
+		if p.Status.ContainerStatuses[0].RestartCount > 1 {
+			for _, condition := range p.Status.Conditions {
+				if condition.Type == "Ready" {
+					if p.Status.Phase != "Running" || condition.Status == "False" {
```
nishantmonu51, line 362: can you add a inline code comment around here, explaining the logic around the if checks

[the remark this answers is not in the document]
AdheipSingh: image

Here if you see is the above image is the exact status for crashloopbackOff
1 . incase you are referring to get a simple way to match the condition, then i could go this route
```
 if  if.Status.ContainerStatuses[0].state.waiting.reason == "CrashLoopBackOff" {
                sdkdelete()
}
```
- Why i dint go with the above is due to mainly these reasons : 

1. The ```waiting.reason``` is a string() , reason can be either ErrorImagePull or ImagePullBackOff  or CrashloopackOff or something else. In all sceanrios sts pod  will hangup and all require deletion manually.  Their can be n cases to match.
Plus pod does not directly go to crashloop it will first through error then go to crashstatus, so this string changes.

 Plus if you see the below image thats a imagePullback 
image

2. Container Restart Count needs to be checked because during initialisation of a pod, the status condition is same, so reconcile loop will keeping on killing it, so in case its > 1 then we match the condition that is it a crashloop ,imagepull, error etc and then delete it. 

3. We are not only dealing with crashloopback, ideally any situation whether error, crashloopback, imagepullback etc which can lead up to hangup state of pod will need to be deleted, with condition.Type and status.Phase we can reach all off them regardless of whatever msg k8s prints out;

AdheipSingh: sure

pkg/controller/druid/handler.go
AdheipSingh: oh i missed this will filter on label ```app=druid```

pkg/controller/druid/handler.go
AdheipSingh: sure

pkg/controller/druid/handler.go
```diff
@@ -169,12 +169,23 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 					return nil
 				}
 
+				if m.Spec.ForcedRollbackSts == true {
+					if nodeSpec.PodManagementPolicy == "OrderedReady" {
```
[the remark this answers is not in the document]
AdheipSingh: sure

docs/upgrade.md
AdheipSingh: sure will do

pkg/controller/druid/handler.go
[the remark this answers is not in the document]
AdheipSingh: > not sure if checking restartcount is the best idea, I wonder if the count can be greater than 1 if the pod is evicted from a node and moved to other node or a node dies etc.

In all these cases even if your restart count is > 1 , the pod will not be deleted cause the sub-condition of type and phase will not match, since pod will be in running state.

pkg/controller/druid/handler.go
himanshug: I think we already add following labels to all the pods...

```
	labels["app"] = "druid"
	labels["druid_cr"] = clusterName
	labels["nodeSpecUniqueStr"] = nodeSpecUniqueStr
```

above labels combined with namespace should identify all pods for a specific statefulset/deployment operator created.

nishantmonu51 (approved):
👍

pkg/controller/druid/handler.go
AdheipSingh: yeah i added makelabels()  :)

docs/upgrade.md
```diff
@@ -4,3 +4,5 @@
...
+- During upgrade if sts is set to ordered ready, the sts controller will not recover from crashloopback state. The issues is referenced [here](https://github.com/kubernetes/kubernetes/issues/67250), and here's a reference [doc](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#forced-rollback)
+- How operator solves this is using the ```forcedRollbackSts``` key, if this key is set to ```true``` the operator will delete the sts pod if its in crashloopback state. Example Scenario: During upgrade, user rolls out a faulty configuration causing the historical pod going in crashing state, user rolls out a valid configuration, the new configuration will not be applied unless user manual delete pods, so solve this scenario operator shall delete the pod automatically without user intervention. ```NOTE: User must be aware of this feature, there might be cases where crashloopback might be caused due probe failure, fault image etc, the operator shall keep on deleting on each re-concile loop. Default Behavior is False ```
```
himanshug, line 8: I think default behavior should be true.

Also, we are not doing any rolling back of the sts and the flag name says otherwise, we can probably call it `forceDeleteStsPodOnError`

AdheipSingh: sure ill change the flag name

BTW i feel default behavior should be false, since this a specific scenario and requires user to be aware of this, thats why i added this in notes
```
User must be aware of this feature, their might be cases where crashloopback might be caused due probe failure, fault image etc, the operator shall keep on deleting on each re-concile loop
```

himanshug: I understand that but there is still the assumption that it can't have a negative effect or else why would anyone able enable it?

or do you think there could be adverse effects of deleting pods selected by our intricate set of checks in `checkCrashStatus` ?

AdheipSingh: @himanshug have done the changes, keeping default to true.

>  do you think there could be adverse effects of deleting pods selected by our intricate set of checks in checkCrashStatus

the only thing here to be noted is, whenever your pod goes into ERROR, CRASHLOOP, IMAGEPULLBACK, operator will keep on deleting the pod. and thats what we want ideally. i m fine keeping it to true

PR #97, opened 2020-12-01
Fix76: Scale STS / PVC vertically to support size resizing of druid nodes

### LINKED ISSUE #76: Operator Should Re-Create Sts when k8s API blocks Sts updates
- So most common issue is when i need to expand sts volume for Historical ( or even MM ), as of now here's how its done:

1. apply new druid-cr, operator can't re-concile new configs since k8s does not allow mutating anything accept podSpec.
2. So i delete sts using --cascad=false to make sure pod doesnt get deleted and  only sts gets deleted.
3. Operator on next re-conciles creates the new sts with new configs.
4. I manually expand pvc ( sc supports volume expansion )

What i would like to automate here, is the second step. I can easily delete sts using operator with cascade, what i am not sure at what events will the operator know that k8s api will not update the sts. If we can get this, this will be a big help.

@himanshug any thoughts what can help us trigger this command ?

AdheipSingh:
Fixes #76

### Sequence Flow
```
- For every node in array of nodes do the following if the node is statefulset
1. Get sts set
2. Get pvc referenced by that sts on label
3. GetDesiredPVCSize from CR
4. GetCurrentPVCSize from sts 
5. Validate the request, shrinkage of pvc not possible, so currentSize cannot be greater than desiredSize, reject request here
5. if GetDesiredPVCSize != GetCurrentPVCSize {
    then delete sts with cascade false
    }
6. Operator on reconcile created the sts, since it sees their is no sts present
7. GetPVCSizeofSts from pvc
8. If GetDesiredPVCSize != GetPVCSizeofSts {
      using deepcopy patch the pvc, dont delete the pvc, no use of update CRUD just patch 
   }
```
### Know Issues
- This function to be called before STS Creation  starts, so on a new druid where no sts is created this function will log saying it cannot find any sts for that nodeType ( cold, hot, mm etc )
- The program does not return here, continues to other flow, after sts is created on next reconcile this error should go away
- 10 sec reconcile being a bit fast, so at times patch may take few more sec, so their might be a error log saying sts could not be updated, but goes away quickly

## TODO:
- [x]  the operator should not allow shrinkage of pvc claims, this should be validated by the operator and logged/sendEvents.
Here's a sample log which shall be thrown
```
{"level":"error","ts":1606744048.5905652,"logger":"druid_operator_handler","msg":"Failed to Patch [PVC:cold-data-master-in-druid-historicals-cold-0] due to [PersistentVolumeClaim \"cold-data-master-in-druid-historicals-cold-0\" is invalid: spec.resources.requests.storage: Forbidden: field can not be less than previous value]","error":"Failed to Patch [PVC:cold-data-master-in-druid-historicals-cold-0] due to [PersistentVolumeClaim \"cold-data-master-in-druid-historicals-cold-0\" is invalid: spec.resources.requests.storage: Forbidden: field can not be less than previous value]"
```
- [x] To validate that no other change apart from pvc size in condition

<hr>

This PR has:

<hr>

 * `handler.go`

@himanshug @nishantmonu51


AdheipSingh:
Here is a sequence of logs 
```
{"level":"info","ts":1606914406.0898578,"logger":"druid_operator_handler","msg":"Detected Change in VolumeClaimTemplate Sizes for Statefuleset [master-in-druid-historicals-cold] in Namespace [master-in], desVolumeClaimTemplateSize: [74Gi], currVolumeClaimTemplateSize: [73Gi]\n, deleteing STS [master-in-druid-historicals-cold] with casacde=false]"}
{"level":"info","ts":1606914407.2999537,"logger":"druid_operator_handler","msg":"[StatefuleSet:master-in-druid-historicals-cold] successfully deleted with casacde=false","name":"master-in","namespace":"master-in"}
{"level":"info","ts":1606914407.3025851,"logger":"druid_operator_handler","msg":"[PVC:cold-data-master-in-druid-historicals-cold-0] successfully Patched with [Size:73Gi]","name":"master-in","namespace":"master-in"}
{"level":"error","ts":1606914408.527233,"logger":"druid_operator_handler","msg":"Failed to update [StatefulSet:master-in-druid-historicals-cold] due to [StatefulSet.apps \"master-in-druid-historicals-cold\" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden]
```
So the last log is fine, since it takes a bit time for reconcile to run and create new sts with the desired spec.

pkg/controller/druid/handler.go
```diff
@@ -390,6 +394,151 @@ func checkCrashStatus(sdk client.Client, m *v1alpha1.Druid) {
...
+			e := fmt.Errorf("StatefulSet [%s] in namespace is invalid [%s],  Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden. ", sts.Name, sts.Namespace)
+			sendEvent(sdk, drd, v1.EventTypeWarning, "UPDATE_FAIL", e.Error())
+			logger.Error(e, e.Error(), "name", drd.Name, "namespace", drd.Namespace)
+			return false
+		}
+	*/
+
+	return true
+}
+
```
AdheipSingh, line 541: @himanshug i have a few question here, when we want to make sure everythingelseissame(), that means
1. metadata should be same ( labels + annotations )
2. podmanagementpolicy should be same
3. Each field inside volumeclaimtemplates should be same
The operator adds labels such as component etc, in that case nodeSpec.Podlables != sts.Labels , this condition will always be true......
any better approach/simpler approach to check, else the only options seems to check each field ?

AdheipSingh: @himanshug do have a look at this query, kind of confused for this function. The rest functions have been tested out.

himanshug: haven't looked in detail, but would something like following work?

say, you have `oldSts` and `newSts` ... and you wanted to check everything else except `VolumeClaimTemplates` is same ... then do

```
oldSts.Spec.VolumeClaimTemplates = newSts.Spec.VolumeClaimTemplates

is reflect.DeepEquals(oldSts, newSts) ?
```

AdheipSingh:
image

AdheipSingh:
@himanshug can you pls review this

AdheipSingh:
https://travis-ci.org/github/druid-io/druid-operator/builds/752248477#L234 
rate limit :exploding_head:

controllers/druid/handler.go
```diff
@@ -391,6 +395,158 @@ func checkCrashStatus(sdk client.Client, m *v1alpha1.Druid) {
...
+			} else {
+				msg := fmt.Sprintf("[PVC:%s] successfully Patched with [Size:%s]", pvc[i].Name, pvcSize.String())
+				logger.Info(msg, "name", drd.Name, "namespace", drd.Namespace)
+				sendEvent(sdk, drd, v1.EventTypeNormal, "PATCH_SUCCESS", msg)
+			}
+		}
+	}
+	// after deleting the sts, it takes some time for the sts to come
+	time.Sleep(10 * time.Second)
+	return
```
AdheipSingh, line 469: - i wonder if this hack is the ideal way to do it, or we can keep this an an env ?

himanshug:
just updating that I am aware of this one ... I did take a quick look today and I maybe wrong but it felt somewhat complex on first look for handling a very special case. I couldn't spend enough time to think through about how best to simplify it though or if that is even possible.

AdheipSingh:
> just updating that I am aware of this one ... I did take a quick look today and I maybe wrong but it felt somewhat complex on first look for handling a very special case. I couldn't spend enough time to think through about how best to simplify it though or if that is even possible.

:) 
i am pretty sure this is the best way to upgrade sts, we have been using this in-house and haven't faced any downtime or disruptions during updates. The only function which i am not using is the ```checkIfStsHasEverythingElseSame``` since for me i really dont care if anything is updated or not. So thats why i requested your review/suggestion on the best way to check that ( open to change )

The only hack which i used is to sleep , this was just used so that the operator does not throw an error log since after deletion it takes some time to upgrade, regardless of the error log their is no issue, just wanted to keep logs clean :) 

```time.Sleep(10 * time.Second)```

ill be happy to explain any issues/queries

**Comment by AdheipSingh** [comment]
@himanshug any update on this here ?

File: controllers/druid/handler.go [hunk 3]
```diff
@@ -391,6 +395,158 @@ func checkCrashStatus(sdk client.Client, m *v1alpha1.Druid) {
...
+			sendEvent(sdk, drd, v1.EventTypeNormal, "DELETE_SUCCESS", msg)
+		}
+	}
+
+	// In case size dont match, patch the pvc with the desiredsize from druid CR
+	if desVolumeClaimTemplateSize.String() != pvcSize.String() {
+		for i := range pvc {
+			patch := client.MergeFrom(pvc[i].DeepCopy())
+			pvc[i].Spec.Resources.Requests[v1.ResourceStorage] = desVolumeClaimTemplateSize
+			err := sdk.Patch(context.TODO(), pvc[i], patch)
```
himanshug, line 454: my understanding is that not all `StorageClass` support the dynamic expansion but only ones for which `allowVolumeExpansion: true` . So, if the storage class used here did not allow volume expansion , what happens then ... should/can we check that volume expansion is allowed and only then try to follow the expansion code path?

controllers/druid/handler.go
```diff
@@ -391,6 +395,158 @@ func checkCrashStatus(sdk client.Client, m *v1alpha1.Druid) {
...
+				sendEvent(sdk, drd, v1.EventTypeWarning, "DELETE_FAIL", e.Error())
+				return
+			} else {
+				msg := fmt.Sprintf("[PVC:%s] successfully Patched with [Size:%s]", pvc[i].Name, pvcSize.String())
+				logger.Info(msg, "name", drd.Name, "namespace", drd.Namespace)
+				sendEvent(sdk, drd, v1.EventTypeNormal, "PATCH_SUCCESS", msg)
+			}
+		}
+	}
+	// after deleting the sts, it takes some time for the sts to come
```
himanshug, line 467: do you mean following?  we just deleted the sts, it can't come up by itself... no ?

```suggestion
	// after updating the pvc, it takes some time for the volume expansion to happen
```

controllers/druid/handler.go
himanshug: that said, I "think" it shouldn't matter ... did you face any problems without the `sleep` here ? My guess is that this sleep isn't needed.

controllers/druid/handler.go
```diff
@@ -391,6 +395,158 @@ func checkCrashStatus(sdk client.Client, m *v1alpha1.Druid) {
...
+	}
+
+	// In case size dont match and dessize > currsize, delete the sts using casacde=false / propagation policy set to orphan
+	// The operator on next reconcile shall create the sts with latest changes
+	if desVolumeClaimTemplateSize.String() != currVolumeClaimTemplateSize.String() {
+		msg := fmt.Sprintf("Detected Change in VolumeClaimTemplate Sizes for Statefuleset [%s] in Namespace [%s], desVolumeClaimTemplateSize: [%s], currVolumeClaimTemplateSize: [%s]\n, deleteing STS [%s] with casacde=false]", sts.Name, sts.Namespace, desVolumeClaimTemplateSize.String(), currVolumeClaimTemplateSize.String(), sts.Name)
+		logger.Info(msg)
+		sendEvent(sdk, drd, v1.EventTypeNormal, "STS_SIZE_DETECTED", msg)
+
+		if err := sdk.Delete(context.TODO(), sts, client.PropagationPolicy(metav1.DeletePropagationOrphan)); err != nil {
```
himanshug, line 438: when we re-create the sts, would the new sts be able to use the existing pods or will that delete-and-recreate all pods?

controllers/druid/handler.go
```diff
@@ -391,6 +395,158 @@ func checkCrashStatus(sdk client.Client, m *v1alpha1.Druid) {
...
+			sendEvent(sdk, drd, v1.EventTypeWarning, "DELETE_FAIL", e.Error())
+		} else {
+			msg := fmt.Sprintf("[StatefuleSet:%s] successfully deleted with casacde=false", sts.Name)
+			logger.Info(msg, "name", drd.Name, "namespace", drd.Namespace)
+			sendEvent(sdk, drd, v1.EventTypeNormal, "DELETE_SUCCESS", msg)
+		}
+	}
+
+	// In case size dont match, patch the pvc with the desiredsize from druid CR
+	if desVolumeClaimTemplateSize.String() != pvcSize.String() {
```
himanshug, line 450: what happens if we just recreated the sts with new size in pvc and did not explicitly patched the PVCs  i.e. what happens if we removed this whole if clause altogether?

controllers/druid/handler.go
```diff
@@ -391,6 +395,158 @@ func checkCrashStatus(sdk client.Client, m *v1alpha1.Druid) {
...
+	if !ok {
+		return
+	}
+
+	// Validate Request, shrinking of pvc not supported
+	// desired size cant be less than current size
+	// in that case re-create sts/pvc which is a user execute manual step
+	if desVolumeClaimTemplateSize.String() < currVolumeClaimTemplateSize.String() {
+		msg := fmt.Sprintf("Request for Shrinking of sts pvc size [sts:%s] in [namespace:%s] is not Supported", sts.Name, sts.Namespace)
+		logger.Info(msg)
```
himanshug, line 426: this should be `ERROR` as user is not expected to try and downscale the PVC size.

controllers/druid/handler.go
```diff
@@ -391,6 +395,158 @@ func checkCrashStatus(sdk client.Client, m *v1alpha1.Druid) {
...
+		sendEvent(sdk, drd, v1.EventTypeWarning, "STS_SIZE_FAIL", msg)
```
himanshug, line 427: STS_PVC_RESIZE_FAIL ?

controllers/druid/handler.go
```diff
@@ -391,6 +395,158 @@ func checkCrashStatus(sdk client.Client, m *v1alpha1.Druid) {
...
+// scalePVCForSts shall scale validate pvc scaling, delete sts using orphan and patch pvc.
+func scalePVCForSts(sdk client.Client, drd *v1alpha1.Druid, ls map[string]string, configMapSHA, serviceName, nodeSpecUniqueStr string, nodeSpec *v1alpha1.DruidNodeSpec) {
+
+	sts := makeStatefulSetEmptyObj()
+	// During creation/initalization of druid cluster, this will fail since the sts will not be present
+	// So logging this as info, event type normal, subsequent reconciles should work fine.
+	if err := sdk.Get(context.TODO(), *namespacedName(nodeSpecUniqueStr, drd.Namespace), sts); err != nil {
+		msg := fmt.Sprintf("failed to get [StatefuleSet:%s] due to [%s]", nodeSpecUniqueStr, err.Error())
+		logger.Info(msg, "name", drd.Name, "namespace", drd.Namespace)
+		sendEvent(sdk, drd, v1.EventTypeNormal, "GET_FAIL", err.Error())
```
himanshug, line 407: so when a brand new druid cluster is created, we will create a whole bunch of GET_FAIL events ? I think, from the `err` we can check if it is caused by non-existence of sts and in that case we shouldn't log/event anything.

controllers/druid/handler.go
AdheipSingh: their no deletion or re-creation of pods, the sts just orphans the pods, on the next reconcilation the new sts is created. New sts uses the same pods.

controllers/druid/handler.go
```diff
@@ -391,6 +395,158 @@ func checkCrashStatus(sdk client.Client, m *v1alpha1.Druid) {
...
+
+	for i := range sts.Spec.VolumeClaimTemplates {
+		currVolumeClaimTemplateSize = sts.Spec.VolumeClaimTemplates[i].Spec.Resources.Requests[v1.ResourceStorage]
+	}
+
+	for i := range pvc {
+		pvcSize = pvc[i].Spec.Resources.Requests[v1.ResourceStorage]
+	}
+
+	return desVolumeClaimTemplateSize, currVolumeClaimTemplateSize, pvcSize
```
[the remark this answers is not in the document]
AdheipSingh: > it seems you are assuming that there is only one pvc in the nodeSpec.VolumeClaimTemplates array ?
yes


```nodeSpec.VolumeClaimTemplates``` is a single pvc, since our goal here specifically to increase the size of sts referenced pvc's ie those pvc, in case other pvc's are mounted on sts they wont be specified in volumeClaimTemplates. Those will be mounted on volumeMounts and Volumes specifying pvc claims there. something like this. 
```
volumes:
- name: mysql-data
  persistentVolumeClaim:
    claimName: mysql-data-pod-name-0.  
```

File: controllers/druid/handler.go (hunk 3 above)
Comment (AdheipSingh) [reply controllers/druid/handler.go:454]: make sense, ill do that, i think most cloud has default enabled for expansion, regardless will do 
thanks for pointing

File: controllers/druid/handler.go (hunk 6 above)
Comment (AdheipSingh) [reply controllers/druid/handler.go:450]: Not patching pvc will still result in the same size as earlier size. In case of no patch, sts volumeclaimtemplate sizes  and pvc sizes will differ.

File: controllers/druid/handler.go (hunk 10 above)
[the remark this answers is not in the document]
Comment (himanshug) [reply controllers/druid/handler.go:519]: `nodeSpec.VolumeClaimTemplates` is an array and user can totally specify two/more pvc templates in there.

File: controllers/druid/handler.go (hunk 5 above)
Comment (himanshug) [reply controllers/druid/handler.go:439]: ok.

File: controllers/druid/handler.go [hunk 11]
```diff
@@ -547,6 +553,130 @@ func checkCrashStatus(sdk client.Client, m *v1alpha1.Druid) {
...
+				sendEvent(sdk, drd, v1.EventTypeWarning, "DELETE_FAIL", e.Error())
+			} else {
+				msg := fmt.Sprintf("[StatefuleSet:%s] successfully deleted with casacde=false", sts.Name)
+				logger.Info(msg, "name", drd.Name, "namespace", drd.Namespace)
+				sendEvent(sdk, drd, v1.EventTypeNormal, "DELETE_SUCCESS", msg)
+			}
+		}
+
+		// In case size dont match, patch the pvc with the desiredsize from druid CR
+		if desVolumeClaimTemplateSize[i].String() != pvcSize[i].String() {
```
AdheipSingh, line 614: Possibilities of a panic here, the array count in desVolumeClaimTemplateSize can be greater than > pvc Size count, which can cause the for loop iteration to fail. On the other hand, array count of pvcSize can be greater than desVolumeClaimTemplateSize since one sts can have multiple replicas each having a pvc. So ideally iteration to take place separately for pvcSize and the index. 
There should be a check if len(pvcSize) <  len(desVolumeClaimTemplateSize) return

controllers/druid/handler.go
```diff
@@ -190,6 +192,19 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
...
+			//  Ignore for the first iteration ie cluster creation, else get sts shall unnecessary log errors.
+
+			if m.Generation > 1 {
+				if isVolumeExpansionEnabled(sdk, m, &nodeSpec) {
+					err := scalePVCForSts(sdk, &nodeSpec, nodeSpecUniqueStr, m)
+					if err != nil {
+						return err
+					}
+				}
+			}
```
AdheipSingh, line 206: Ill add a flag for this, users should be aware of this feature.

AdheipSingh:
@himanshug over to you for review.

controllers/druid/handler.go
nishantmonu51: is the flag added now ?

nishantmonu51 (approved):

controllers/druid/handler.go
```diff
@@ -728,6 +743,153 @@ func isObjFullyDeployed(sdk client.Client, nodeSpec v1alpha1.DruidNodeSpec, node
...
+	// we would like the operator to create sts.
+	sts, err := readers.Get(context.TODO(), sdk, nodeSpecUniqueStr, drd, func() object { return makeStatefulSetEmptyObj() }, emitEvent)
+	if err != nil {
+		return nil
+	}
+
+	pvcLabels := map[string]string{
+		"druid_cr": drd.Name,
+	}
+
```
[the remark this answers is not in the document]
AdheipSingh: label component for sts/deploy > pods > pvc's

PR #101, opened 2020-12-03
Fix: pvc provisioned by operator

AdheipSingh:
https://github.com/druid-io/druid-operator/pull/70
As discussed in the above PR, operator provisions pvcs, user can specify using volume mounts

once approved/discussed will add documentation

### Description

<hr>

This PR has:

<hr>

 * `types.go`
* `handler.go`

### DIFF
--- a/pkg/apis/druid/v1alpha1/druid_types.go
+++ b/pkg/apis/druid/v1alpha1/druid_types.go
@@ -249,6 +249,9 @@ type DruidNodeSpec struct {
 	// Optional
 	Lifecycle *v1.Lifecycle `json:"lifecycle,omitempty"`
 
+	// Optional: Persitance Volume Spec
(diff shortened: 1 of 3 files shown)


pkg/apis/druid/v1alpha1/druid_types.go
```diff
@@ -249,6 +249,9 @@ type DruidNodeSpec struct {
 	// Optional
 	Lifecycle *v1.Lifecycle `json:"lifecycle,omitempty"`
 
+	// Optional: Persitance Volume Spec
+	PersistentVolumeClaim *v1.PersistentVolumeClaimSpec `json:"persistentVolumeClaimSpec,omitempty"`
```
himanshug, line 253: can we support a whole array of those ?

PR #102, opened 2020-12-03
Fix: 96 Reconcile As Env

### LINKED ISSUE #96: Reconcile Loop time as ENV
- Can we use an ENV for reconcile loop time, we can keep 10 sec to default
- https://github.com/druid-io/druid-operator/blob/master/pkg/controller/druid/druid_controller.go#L103 
- Their are cases where we want to lower the reconcile time etc, its more flexible in an env :)

AdheipSingh:
Fixes #96 

Need to add this in docs 

### Description

<hr>

This PR has:

<hr>

 * `druid_controller.go`
 * `operator.yaml`

### DIFF
--- a/pkg/controller/druid/druid_controller.go
+++ b/pkg/controller/druid/druid_controller.go
@@ -2,6 +2,7 @@ package druid
 
 import (
 	"context"
+	"os"
 	"time"
 
 	druidv1alpha1 "github.com/druid-io/druid-operator/pkg/apis/druid/v1alpha1"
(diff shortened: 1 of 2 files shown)


pkg/controller/druid/druid_controller.go
```diff
@@ -100,6 +101,22 @@ func (r *ReconcileDruid) Reconcile(request reconcile.Request) (reconcile.Result,
...
 	} else {
-		return reconcile.Result{RequeueAfter: time.Second * 10}, nil
+		return reconcile.Result{RequeueAfter: lookupReconcileTime()}, nil
+	}
+}
+
+// lookupReconcileTime shall return reconcile time if set as ENV
+// else defaults to 10s
+func lookupReconcileTime() time.Duration {
+	val, exists :=  os.LookupEnv("RECONCILE_TIME")
```
himanshug, line 111: nit: RECONCILE_WAIT_ON_ERROR ?


also, can we parse that on startup and fail the operator startup if env variable was provided but invalid?

AdheipSingh: sure make sense
But why WAIT ON ERROR ? curious :thinking:

AdheipSingh:
since reconcile method attached to reconcileDruid struct and this will be initalized in main.go, i added this field to the struct itself, so it'll be called automatically during start on controller.
Plus i feel its better to hook such fields to the struct itself for better understanding/readability.
Have tested this in dev, seems fine. 

Once approved ill add a few lines in the doc @himanshug

pkg/controller/druid/druid_controller.go
himanshug: I re-read this code a bit and realized that env is actually used for non-error case, `RECONCILE_WAIT` was the correct term, my mistake.

AdheipSingh: ah ha i have been using that in my local :), will send the change and doc too. @himanshug

PR #104, opened 2020-12-04
Fix #103: missing labels in deploy,sts

### LINKED ISSUE #103: Deployments/Sts missing Labels
- Labels are only being passed on to pods, deployments and sts created by the operator do not have labels in their metadata.
- DD/Prometheus monitoring based on kube metrics, filters deployments/sts on basis of labels. Few metrics arn't being able to show up due to missing labels.
- We can pass the same set of labels to deployments and sts if defined in CR else default ones should be passed ie component, app etc.

AdheipSingh:
Fixes #103 

### Description

<hr>

This PR has:

<hr>

 * `handler.go`


AdheipSingh:
@himanshug in case you missed this

PR #113, opened 2020-12-13
add a smoke test that deploy operator, a cr and validates reconciliation using controller-runtime envtest package

himanshug:
note that envtest ( https://godoc.org/sigs.k8s.io/controller-runtime/pkg/envtest ) brings up only the k8s API Server and not all the usual controllers but still good enough to test that operator was able to create necessary resources it was supposed to and submitted them to API server.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,22 +2,22 @@
 FROM golang:1.13 as builder
 
 WORKDIR /workspace
-# Copy the Go Modules manifests
-COPY go.mod go.mod
-COPY go.sum go.sum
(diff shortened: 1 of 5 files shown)


AdheipSingh:
Intresting I wonder if we can run it against an api and run integration test as you were referring in the apache druid repo.

himanshug:
> Intresting I wonder if we can run it against an api and run integration test as you were referring in the apache druid repo.

yes, we will do that as well.... this is just a simpler version which is easy to setup and doing most of the validation of what operator is supposed to do.

PR #114, opened 2020-12-14
add OpenAPI schema validation to Druid CR

himanshug:
### Description

updated things to generate Druid CR schema validation, however that leads to a bloated CRD spec with 7000+ lines ... gonna look a bit more into reducing its size or other options.

<hr>

This PR has:

<hr>


AdheipSingh:
ah ha , i was thinking of adding the webhook since its scaffolded  with kube builder :)

AdheipSingh:
Does this mean can we remove the validation block in the handler ?

deploy/crds/druid.apache.org_druids_crd.yaml
```diff
@@ -1,43 +0,0 @@
...
-        status:
-          description: DruidStatus defines the observed state of Druid
-          type: object
-      type: object
-  version: v1alpha1
-  versions:
-  - name: v1alpha1
-    served: true
-    storage: true
-
```
AdheipSingh, line 43: is this removing the CRD ?

AdheipSingh: oh i see its in diff file, i wonder  where are we adding the kubebuilder tags for validation

himanshug:
adding webhook is easy actually since we already have a `verifyDruidSpec(cr)` defined ... however webhook deployment requires TLS certs which is additional work for k8s cluster operator specially if they don't need those for any other use cases.. so webhook would be kept optional anyways.

openapi schema validation is good as it needs nothing extra from the user's side but I just need to overcome those 7000+ lines, since they are generated so it is ok probably but wanna take a look at what is the best practice there.

AdheipSingh:
yeah we will need to add tags, on the struct 
something like this
```
// +k8s:openapi-gen=true
type druidNodeSpec struct {
	// +kubebuilder:validation:Minimum=1
	Replicas int32 `json:"replicas,omitempty"
```

AdheipSingh:
yeah webhook will address larger usecases, i dont have any as of now, since i just need to have valid validation at the API level.
But like the recent vunerability detected in k8s was patching of external ips in k8s service, we could add a validation for that on webhook and support it in operator. ( https://groups.google.com/g/kubernetes-security-announce/c/iZWsF9nbKE8?pli=1 )

himanshug:
I did some looking around and it seems large crd files are common if one wants schema validation at this time, so it is ok I guess e.g.

https://github.com/prometheus-operator/prometheus-operator/blob/548a3179e3439046418ddbf452338e423b035b6c/example/prometheus-operator-crd/monitoring.coreos.com_prometheuses.yaml

https://github.com/strimzi/strimzi-kafka-operator/blob/c440bcec056ad5a664b861de5abb464f4863cda8/install/cluster-operator/040-Crd-kafka.yaml

AdheipSingh:
@himanshug when merge kindly squash the commits, its easy to cherry pick and keep track of changes :)

himanshug:
looks like we are hitting docker pull limits ...

```
1.13: Pulling from library/golang
anyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
The command "docker build -t druidio/druid-operator:testversion ." exited with 1.
```

AdheipSingh:
> ```
> // +k8s:openapi-gen=true
> type druidNodeSpec struct {
> 	// +kubebuilder:validation:Minimum=1
> 	Replicas int32 `json:"replicas,omitempty"
> ```
@himanshug i dont see any such validation, dint we need to add tags to validate the spec ?

himanshug:
@AdheipSingh https://github.com/druid-io/druid-operator/blob/master/apis/druid/v1alpha1/druid_types.go#L150 , just kept that to 0 based on previous discussion that autoscaling sometime can set the replica count to 0.

AdheipSingh:
@himanshug i am thinking of adding more tags for required keys.
we can also comment off https://github.com/druid-io/druid-operator/blob/master/controllers/druid/handler.go#L1249 
sounds good ?

himanshug:
> i am thinking of adding more tags for required keys.

default is required unless json tag contains 'omitempty' ... I updated json tags appropriately . but feel free to update if I missed something.

> we can also comment off https://github.com/druid-io/druid-operator/blob/master/controllers/druid/handler.go#L1249

I would still keep that as backup of things either not caught by open api validation or for users that might want to use crd version without open api validation schema in it.

PR #115, opened 2020-12-16
reconcile sts/deploy if replica count changed

### LINKED ISSUE #108: Watch deployments/Sts/services with watcher
- The operator is only watching Druid CR, which is fine, but if someone directly edits the deployment, it should make sure the desiredState remains with CR. Until and update is run it does not change the state.

- Druid Operator should watch for deployments/sts/services/pdb etc resources which it creates and make sure their state is met.

@himanshug any specific reason for commenting out the watchers for deployments/sts etc ? i see a TODO on the comment though :) 
https://github.com/druid-io/druid-operator/blob/master/pkg/controller/druid/druid_controller.go#L56

himanshug:
Fixes #108

### Description

This PR introduces the mechanism to do comparison of resources generated by druid-operator so as to reconcile them if they were changed directly. currently it only does so for the Deployment/StatefulSet replica counts but could be extended further as needed.

<hr>

This PR has:
- [] added documentation for new or modified features or behaviors.

<hr>

### DIFF
--- a/controllers/druid/handler.go
+++ b/controllers/druid/handler.go
@@ -84,7 +84,7 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 	if _, err := sdkCreateOrUpdateAsNeeded(sdk,
 		func() (object, error) { return makeCommonConfigMap(m, ls) },
 		func() object { return makeConfigMapEmptyObj() },
-		nil, m, configMapNames); err != nil {
+		alwaysTrueIsEqualsFn, noopUpdaterFn, m, configMapNames); err != nil {
 		return err
 	}
 
@@ -112,7 +112,7 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 		if _, err := sdkCreateOrUpdateAsNeeded(sdk,
 			func() (object, error) { return nodeConfig, nil },
 			func() object { return makeConfigMapEmptyObj() },
-			nil, m, configMapNames); err != nil {
+			alwaysTrueIsEqualsFn, noopUpdaterFn, m, configMapNames); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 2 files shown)


controllers/druid/handler.go
```diff
@@ -831,6 +844,13 @@ func makeStatefulSet(nodeSpec *v1alpha1.DruidNodeSpec, m *v1alpha1.Druid, ls map
 	}, nil
 }
 
+// currently only checks for replica count mismatch, can be extended further
+func statefulSetIsEquals(obj1, obj2 object) bool {
+	o1 := obj1.(*appsv1.StatefulSet)
+	o2 := obj2.(*appsv1.StatefulSet)
+	return *o1.Spec.Replicas == *o2.Spec.Replicas
+}
+
```
AdheipSingh, line 853: @himanshug in case we add spec here, wonder if this will  conflict on scaling sts https://github.com/druid-io/druid-operator/pull/97 

Regardless are you planning to keep only replicas for now or extend for spec too

AdheipSingh:
question :- 
1. What will happen if someone delete's a deployment/sts, with this PR will operator create the sts/deploy ?

controllers/druid/handler.go
himanshug: for now, I am leaving it at just replica count mismatch ... but we can add more stuff here based on real world needs.

himanshug:
> What will happen if someone delete's a deployment/sts, with this PR will operator create the sts/deploy ?

operator would recreate those in 10secs in the worst case.

connie-ru-wang:
Hi, is it possible that this PR broke the HPA functionality? I have a working hpa in a cluster with operator:0.0.4, but the hpa doesn't work in a cluster with operator:0.0.5 because it looks like the operator immediately reconciles the sts size back down when the hpa tries to scale up.

AdheipSingh:
@connie-ru-wang 
yeah i think that might have broken the HPA, we can add a condition to check in case HPA is enabled, or maybe add a flag for force maintaining the replica count at the operator level. 
Regardless do you mind telling for which components are you using HPA for ? 

cc @himanshug

AdheipSingh:
BTW @connie-ru-wang  it would be nice if you can create a separate issue for this, so that it can be tracked. 
Thanks

connie-ru-wang:
@AdheipSingh I've been testing HPA with routers, brokers, historicals. Although for historicals, most likely we will need v2beta2 support for the specific behavior policies that will prevent thrashing.  https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#support-for-configurable-scaling-behavior

PR #122, opened 2021-01-03
Update controller-gen to v0.4.1, regenerate CRD to support Kubernetes…

### LINKED ISSUE #121: CRD Problem with Kubernetes Versions >= 1.18
On a Kubernetes 1.18 cluster (Amazon EKS)

```bash
kubectl apply -f deploy/crds/druid.apache.org_druids.yaml

The CustomResourceDefinition "druids.druid.apache.org" is invalid: 
* spec.validation.openAPIV3Schema.properties[spec].properties[services].items.properties[spec].properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[nodes].additionalProperties.properties[services].items.properties[spec].properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
```

Some internets searches led me to what looks like the same issue on a different project:
https://github.com/prometheus-operator/prometheus-operator/issues/3122

Their solution was https://github.com/prometheus-operator/prometheus-operator/issues/3137:

> As reported in #3122, the latest version of sigs.k8s.io/controller-tools (>= v0.28.0) generate CustomResourceDefinition objects that aren't compatible with Kubernetes >= v1.18.
> 
>The recommended approach is to switch from apiextensions.k8s.io/v1beta1 to apiextensions.k8s.io/v1.

I see that this project is still using CRD version `apiextensions.k8s.io/v1beta1`. I presume the same fix is necessary.

kurtostfeld:
Fixes #121

### Description

Before this change, the included CRD would not work with Kubernetes 1.18 or later.

The only manual change made was bumping the controller-gen version in `Makefile` from `v0.2.5` to `v0.4.1`.

Then I ran `make manifests` to regenerate the CRD yaml from the .go source.

The generated CRD is now version `apiextensions.k8s.io/v1` instead of `apiextensions.k8s.io/v1beta1`, which requires Kubernetes 1.16+.

This PR has:

I can verify that before this change, I got this error on a Kubernetes 1.18 (Amazon EKS) cluster:

```bash
kubectl apply -f deploy/crds/druid.apache.org_druids.yaml

The CustomResourceDefinition "druids.druid.apache.org" is invalid: 
* spec.validation.openAPIV3Schema.properties[spec].properties[services].items.properties[spec].properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[nodes].additionalProperties.properties[services].items.properties[spec].properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
```

after this change, the same command works perfectly on the same cluster:

```bash
kubectl apply -f deploy/crds/druid.apache.org_druids.yaml

customresourcedefinition.apiextensions.k8s.io/druids.druid.apache.org created
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ ifeq (, $(shell which controller-gen))
 	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
 	cd $$CONTROLLER_GEN_TMP_DIR ;\
 	go mod init tmp ;\
-	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.5 ;\
+	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1 ;\
 	rm -rf $$CONTROLLER_GEN_TMP_DIR ;\
 	}
 CONTROLLER_GEN=$(GOBIN)/controller-gen
--- a/deploy/crds/druid.apache.org_druids.yaml
+++ b/deploy/crds/druid.apache.org_druids.yaml
@@ -1,10 +1,10 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.2.5
+    controller-gen.kubebuilder.io/version: v0.4.1
   creationTimestamp: null
   name: druids.druid.apache.org
 spec:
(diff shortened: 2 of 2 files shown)


himanshug:
is this backward compatible [ with 1.16 ] ?

kurtostfeld:
> is this backward compatible [ with 1.16 ] ?

Yes, this should work fine with 1.16+, although I haven't personally tested it. This won't work with earlier versions as v1 CRD wasn't supported.

Also, apparently there is a test failure I need to fix.

himanshug:
build runs an integration test with `Kubernetes v1.16.4` , so if this change is ok with 1.16+ then the build should work I think.

PR #124, opened 2021-01-08
Fix #43 Finalizer for pvc cleanup

### LINKED ISSUE #43: Operator should delete PVC once CR is deleted
- Currently the operator does not remove PVC's after the CR has been removed.

### Proposal
- In the types.go have a volumeReclaimPolicy if set to delete, the operator should clear up the pvc associated with that CR.
- Ideal way is to have a finalizer which shall run up clean up task to remove pvc. 
-------------------------------------------------------------------------------------------------------------------------------
- Additionally the operator should also support cleaning of pvc's which are not mounted or referenced by the CR pods. Statefulset do not delete the pvc associated by the pod. so incase of scale up/down pvc's keep on hanging up. This can also be added optionally.

@himanshug

AdheipSingh:
Fixes #43 
- As discussed in issue, adding finalizer  to clean up pvc.
- pvc will get cleaned only after sts is deleted else pvc finalizer will come into play

### Description

<hr>

This PR has:

<hr>

 * `hander.go`
 * `api.go`
 * `util.go`
@himanshug


apis/druid/v1alpha1/druid_types.go
```diff
@@ -46,6 +43,9 @@ type DruidSpec struct {
 	// Required: in-container directory to mount with common.runtime.properties
 	CommonConfigMountPath string `json:"commonConfigMountPath"`
 
+	// Optional: If set to false, pvc shall be deleted on deletion of CR
+	VolumeReclaimPolicy bool `json:"volumeReclaimPolicy"`
```
himanshug, line 47: thanks, this is great :)

what do you think about having `enablePVCDeletionFinalizer` instead ?  And have separate finalizers for separate things to finalize instead of one `finalizers.druid.apache.org` e.g.
when `enablePVCDeletionFinalizer = TRUE` , then operator would add a finalizer `delete-pvc.finalizers.druid.apache.org`  ... in future we might add more type of finalizers.

also, by default set `enablePVCDeletionFinalizer = TRUE` as druid's data is anyways safe in deep storage.

AdheipSingh: agreed will do that changes

AdheipSingh:
@himanshug in case you missed :)

controllers/druid/handler.go
```diff
@@ -88,6 +89,45 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
...
+		/*
+			If finalizer isn't present add it to object meta.
+			In case cr is already deleted do not call this function
+		*/
+		cr := checkIfCRExists(sdk, m)
+		if cr == true {
+			if !ContainsString(m.ObjectMeta.Finalizers, finalizerName) {
+				m.SetFinalizers(append(m.GetFinalizers(), finalizerName))
+				if err := sdk.Update(context.Background(), m); err != nil {
+					e := fmt.Errorf("failed to Update druid CR for [%s] due to [%s]", m.Name, err.Error())
```
himanshug, line 122:
```suggestion
					e := fmt.Errorf("failed to add finalizer to druid CR for [%s] due to [%s]", m.Name, err.Error())
```

controllers/druid/handler.go
```diff
@@ -88,6 +89,45 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
...
+				return err
+			} else {
+				return nil
+			}
+		}
+		/*
+			If finalizer isn't present add it to object meta.
+			In case cr is already deleted do not call this function
+		*/
+		cr := checkIfCRExists(sdk, m)
```
himanshug, line 117: can we do this check after `ContainsString(m.ObjectMeta.Finalizers, finalizerName)` check so that we don't unnecessarily get the CR if finalizer is already there.

controllers/druid/handler.go
```diff
@@ -336,6 +375,129 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
...
+	sts := make([]*appsv1.StatefulSet, 0)
+	for i := range stsList.Items {
+		sts = append(sts, &stsList.Items[i])
+	}
+
+	return sts, nil
+
+}
+
+func finalizer(sdk client.Client, m *v1alpha1.Druid) error {
```
himanshug, line 471:
```suggestion
func executeFinalizers(sdk client.Client, m *v1alpha1.Druid) error {
```

controllers/druid/handler.go
```diff
@@ -336,6 +375,129 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
...
+		} else {
+			msg := fmt.Sprintf("Finalizer success for CR [%s] in namespace [%s]", m.Name, m.Namespace)
+			sendEvent(sdk, m, v1.EventTypeNormal, "TRIGGER_FINALIZER_SUCCESS", msg)
+			logger.Info(msg)
+		}
+
+		// remove our finalizer from the list and update it.
+		m.ObjectMeta.Finalizers = RemoveString(m.ObjectMeta.Finalizers, finalizerName)
+		if err := sdk.Update(context.TODO(), m); err != nil {
+			e := fmt.Errorf("failed to Update druid CR for [%s] due to [%s]", m.Name, err.Error())
```
himanshug, line 490:
```suggestion
			e := fmt.Errorf("failed to remove finalizer from druid CR for [%s] due to [%s]", m.Name, err.Error())
```

AdheipSingh:
@himanshug thanks for reviewing will get back with updates :)

controllers/druid/handler.go
AdheipSingh: ok so here's what i think
1. we check if CR exists
2. then check if that CR has string present or not
if we do the check after containerstring func that does not seems appropriate to me, since we must confirm if cr is present or removed
i hope thats why you meant right ? regardless open to suggestions :)

himanshug: ok, I see that method is also getting the current spec and not just checking the existence ...

AdheipSingh:
i finalized testing with the latest changes on this @himanshug 
LGTM
image

PR #130, opened 2021-01-21
Add helm chart

dsx:
### Description

This is to add Helm chart. Some work has been started in #33, however for some reason it has been abandoned very quick.

<hr>

This PR has:

<hr>

 * `chart`

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -1,0 +1,23 @@
+apiVersion: v2
+name: druid-operator
+description: Druid Kubernetes Operator
+
+# A chart can be either an 'application' or a 'library' chart.
+#
+# Application charts are a collection of templates that can be packaged into versioned archives
+# to be deployed.
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+appVersion: 0.0.4
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -20,4 +20,4 @@ version: 0.1.0
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 0.0.4
+appVersion: 0.0.5
--- a/chart/templates/crd.yaml
+++ b/chart/templates/crd.yaml
@@ -1,0 +1,45 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: druids.druid.apache.org
+  labels:
+    {{- include "druid-operator.labels" . | nindent 4 }}
+spec:
+  group: druid.apache.org
+  names:
+    kind: Druid
+    listKind: DruidList
+    plural: druids
+    singular: druid
+  scope: Namespaced
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      description: Druid is the Schema for the druids API
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
+            of an object. Servers should convert recognized schemas to the latest
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
+          description: DruidSpec defines the desired state of Druid
+          type: object
+        status:
+          description: DruidStatus defines the observed state of Druid
+          type: object
+      type: object
+  version: v1alpha1
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: true
+    
--- a/chart/templates/crd.yaml
+++ b/chart/templates/crd.yaml
@@ -1,9 +1,10 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
+  annotations:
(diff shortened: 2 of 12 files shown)


chart/templates/deployment.yaml
```diff
@@ -0,0 +1,61 @@
...
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          env:
+            - name: WATCH_NAMESPACE
+              value: {{ .Values.watchNamespace }}
+            - name: DENY_LIST
+              value: {{ .Values.denyList }}
+            - name: RECONCILE_WAIT_ON_ERROR
+              value: {{ .Values.reconcileWaitOnError }}
```
AdheipSingh, line 41: This needs to be ```RECONCILE_WAIT```

chart/values.yaml
```diff
@@ -0,0 +1,67 @@
...
+# Declare variables to be passed into your templates.
+
+# Namespace to watch or empty string to watch all namespaces
+watchNamespace: ""
+
+# Comma-separated list of namespaces to ignore
+denyList: "default,kube-system"
+
+# Reconciliation delay
+reconcileWaitOnError: 10s
```
AdheipSingh, line 12: nit: ```reconcileWait```
Ref:  https://github.com/druid-io/druid-operator/blob/master/controllers/druid/druid_controller.go#L65

chart/templates/deployment.yaml
```diff
@@ -0,0 +1,61 @@
...
+      serviceAccountName: {{ include "druid-operator.serviceAccountName" . }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          env:
```
AdheipSingh, line 35: How about not hard coding values, and keeping this env too 
```
{{- range $key, $value :=  .Values.env }}
        - name: {{ $key }}
          value: {{ tpl $value $ | quote }}
{{- end }}
 ```

File: chart/templates/deployment.yaml (hunk 1 above)
Comment (AdheipSingh) [inline chart/templates/deployment.yaml:41]: can add ```value: {{ .Values.myvalue  | quote }}```

File: chart/values.yaml [hunk 4]
```diff
@@ -0,0 +1,67 @@
...
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+# Namespace to watch or empty string to watch all namespaces
+watchNamespace: ""
+
+# Comma-separated list of namespaces to ignore
+denyList: "default,kube-system"
+
+# Reconciliation delay
```
AdheipSingh, line 11: nit: Reconcilation Time Period

chart/Chart.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v2
```
AdheipSingh, line 1: is this helm api version or charts ?

chart/templates/crd.yaml
```diff
@@ -0,0 +1,45 @@
...
+        status:
+          description: DruidStatus defines the observed state of Druid
+          type: object
+      type: object
+  version: v1alpha1
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: true
+
```
AdheipSingh, line 45: CRD has been updated
Ref: https://github.com/druid-io/druid-operator/blob/master/deploy/crds/druid.apache.org_druids.yaml

chart/Chart.yaml
dsx: This is auto-generated by `helm create`, I don't really know TBH.

chart/templates/deployment.yaml
dsx: I left `POD_NAME` as it was, although I'm not sure if that's still used.

chart/Chart.yaml
AdheipSingh: BTW is this tested on helm v2 or v3 ?

dsx: Helm v3 only. Helm v2 is EOL as far as I understand.

himanshug:
LGTM after @AdheipSingh 's +1

AdheipSingh:
@dsx can you add some docs around this.
helm template, helm update and helm rollback commands would be useful with this chart.

chart/templates/deployment.yaml
```diff
@@ -0,0 +1,57 @@
...
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "druid-operator.serviceAccountName" . }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```
AdheipSingh, line 33: @dsx you are missing command here, with the ```master``` branch ( kubebuilder ) it requires ```manager``` to start.
ref: https://github.com/druid-io/druid-operator/blob/master/deploy/operator.yaml

I understand you tested this with operator-sdk layout yaml but i think we would like to keep this in sync with master branch changes, so in that case i recommend to test this chart with docker image build from master, since a release might be cut somewhat soon.

cc @himanshug

AdheipSingh (changes requested):

AdheipSingh:
@dsx 
can we keep the crd installation separate, what i mean is if a user uninstalls the chart, in the current case the CRD get removed which can affect any cluster already running, this happened to me where recently playing around helm, i accidentally removed a chart which resulted in deleting the CRD and cluster configs and cluster running got vanished. To avoid this i see two options

1. Keep CRD installation/uninstallation totally seperate of helm charts. I also learned this from here https://github.com/jetstack/cert-manager/blob/master/deploy/charts/cert-manager/README.template.md#installing-the-chart
OR
2. Protect the CRD resource by default by adding this too the helm chart 
```
 annotations:
    helm.sh/resource-policy: keep
```
+ cc @himanshug

himanshug:
yeah, sounds reasonable that accidentally removing the CRD might be catastrophic in some cases. I would say, by default,

helm install SHOULD install the CRD (so that new users just get it out of the box)
helm uninstall should NOT remove the CRD (so that CRD doesn't get deleted accidentally)

AdheipSingh:
this is great, LGTM

chart/templates/deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
```
himanshug, line 33: curious, why is AppVersion default value not in values.yaml unlike all other properties e.g. image.tag could have value `0.0.5` by default instead of being empty?

dsx: `.Chart.AppVersion` is chart-specific and needs to be updated on release. It can't be in values.yaml by design.

himanshug: ok, I get it, thanks.

PR #142, opened 2021-02-26
Delete unmounted/orphaned pvc

### LINKED ISSUE #136: Deletion of Unmounted PVC's
### Scenario
- when ingestion kicks of i scale my MM to  10 ( all statefulsets ), now i have 10 pvc for 10 MM Statefulsets
- when done i scale down MM to 1, now i have 1 MM statefulsets and 10 PVC of which 9 pvc are unmounted .

### ISSUE
- These 9 pvc's shall exist till the time the CR is deleted ( where finalizer shall come into play and the hook deletes the pvc's ).

### SOLUTION
- The 9 pvc's are of no / little use when we scale down druid nodes, the operator should ideally remove the pvc which are not mounted by running a check on each reconciliation loop.
- This way all the pvc's which are hanging up are cleaned.

@himanshug

AdheipSingh:
Fixes #136

### Description

- As discussed in the issue, this PR will make sure unmounted PVC's are deleted on each reconcile. 

<hr>

This PR has:

<hr>

 * `handler.go` 
 
 @himanshug

### DIFF
--- a/controllers/druid/handler.go
+++ b/controllers/druid/handler.go
@@ -253,6 +253,12 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 		}
 	}
 
+	if err := deleteOrphanPVC(sdk, m); err != nil {
+		e := fmt.Errorf("Error in deleteOrphanPVC [%s]", err.Error())
+		sendEvent(sdk, m, v1.EventTypeWarning, "LIST_FAIL", e.Error())
+		logger.Error(e, e.Error(), "name", m.Name, "namespace", m.Namespace)
+	}
+
 	//update status and delete unwanted resources
 	updatedStatus := v1alpha1.DruidStatus{}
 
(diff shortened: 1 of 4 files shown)


controllers/druid/handler.go
```diff
@@ -253,6 +253,12 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 		}
 	}
 
+	if err := deleteOrphanPVC(sdk, m); err != nil {
+		e := fmt.Errorf("Error in deleteOrphanPVC [%s]", err.Error())
+		sendEvent(sdk, m, v1.EventTypeWarning, "LIST_FAIL", e.Error())
+		logger.Error(e, e.Error(), "name", m.Name, "namespace", m.Namespace)
+	}
+
```
AdheipSingh, line 261: - We can add a flag in case this wants does not want to be enforced ( may be user wants to keep data in pvc )

himanshug: yeah, that would make sense... default it to true though.

AdheipSingh:
@himanshug

PR #146, opened 2021-03-11
Proposal: Introduce Interfaces for Read/Write Operations

AdheipSingh:
## Proposal

- Introduce two interfaces ( https://github.com/AdheipSingh/druid-operator/blob/interfaces/controllers/druid/interface.go#L13 )
1. Read Interface: Get and List operations
2. Write Interface: Create, Update, Delete and Patch Operations

- Struct to have entities of type func() [ this func() is already present in the handler, just using this as type here ] https://github.com/AdheipSingh/druid-operator/blob/interfaces/controllers/druid/interface.go#L18

- Use Methods instead of funcs, and each method has a receiver to struct funcs. https://github.com/AdheipSingh/druid-operator/blob/interfaces/controllers/druid/interface.go#L25
- Initialise the interface to be used anywhere in the package druid https://github.com/AdheipSingh/druid-operator/blob/interfaces/controllers/druid/interface.go#L23

- This way the benefits i see is:
1. Not write specific func for various kubernetes objects. 
2. Code Clean Up, this will reduce the size of the handler.
3. The underlying methods shall remain immutable, lets say we need to improve logging, we can just change it in the method and it shall be implemented in all the handler since its implemented as an interface.
4. Better understanding, anyone can come in use the interface, great readability of the code.

@himanshug looking forward to your review. I took most of the design from controller-runtime. I don't plan to change any core functions, just give them an interface type and some refactoring of the handler.

The current changes have been tested for the finalizer feature and gave +ve results.

Once i get a +ve feedback, ill add more interfaces and start working on this.

### Description

<hr>

This PR has:

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

### DIFF
--- a/controllers/druid/handler.go
+++ b/controllers/druid/handler.go
@@ -68,6 +68,7 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 	podDisruptionBudgetNames := make(map[string]bool)
 	hpaNames := make(map[string]bool)
 	ingressNames := make(map[string]bool)
+	pvcNames := make(map[string]bool)
 
 	ls := makeLabelsForDruid(m.Name)
 
@@ -247,6 +248,18 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 				return err
 			}
 		}
+
+		if nodeSpec.PersistentVolumeClaim != nil {
+			for _, pvc := range nodeSpec.PersistentVolumeClaim {
+				if _, err := sdkCreateOrUpdateAsNeeded(sdk,
+					func() (object, error) { return makePersistentVolumeClaim(&pvc, &nodeSpec, m, lm, nodeSpecUniqueStr) },
+					func() object { return makePersistentVolumeClaimEmptyObj() }, alwaysTrueIsEqualsFn,
+					noopUpdaterFn,
+					m, pvcNames); err != nil {
+					return err
+				}
+			}
+		}
 	}
 
 	if m.Spec.DeleteOrphanPvc {
@@ -343,6 +356,19 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 			return result
 		})
 	sort.Strings(updatedStatus.ConfigMaps)
+  
+        updatedStatus.PersistentVolumeClaims = deleteUnusedResources(sdk, m, pvcNames, ls,
+		func() runtime.Object { return makePersistentVolumeClaimListEmptyObj() },
+		func(listObj runtime.Object) []object {
+			items := listObj.(*v1.PersistentVolumeClaimList).Items
+			result := make([]object, len(items))
+			for i := 0; i < len(items); i++ {
+				result[i] = &items[i]
+			}
+			return result
+		})
+	sort.Strings(updatedStatus.PersistentVolumeClaims)
+  
 
 	podList, _ := readers.List(context.TODO(), sdk, m, makeLabelsForDruid(m.Name), func() runtime.Object { return makePodList() }, func(listObj runtime.Object) []object {
 		items := listObj.(*v1.PodList).Items
@@ -1195,6 +1221,36 @@ func makeIngress(nodeSpec *v1alpha1.DruidNodeSpec, m *v1alpha1.Druid, ls map[str
 	return ingress, nil
 }
 
+func makePersistentVolumeClaim(pvc *v1.PersistentVolumeClaim, nodeSpec *v1alpha1.DruidNodeSpec, m *v1alpha1.Druid, ls map[string]string, nodeSpecUniqueStr string) (*v1.PersistentVolumeClaim, error) {
+
+	pvc.TypeMeta = metav1.TypeMeta{
+		APIVersion: "v1",
+		Kind:       "PersistentVolumeClaim",
+	}
+
+	pvc.ObjectMeta.Namespace = m.Namespace
+
+	if pvc.ObjectMeta.Labels == nil {
+		pvc.ObjectMeta.Labels = ls
+	} else {
+		for k, v := range ls {
+			pvc.ObjectMeta.Labels[k] = v
+		}
+	}
+
+	if pvc.ObjectMeta.Name == "" {
+		pvc.ObjectMeta.Name = nodeSpecUniqueStr
+	} else {
+		for _, p := range nodeSpec.PersistentVolumeClaim {
+			pvc.ObjectMeta.Name = p.Name
+			pvc.Spec = p.Spec
+		}
+
+	}
+
+	return pvc, nil
+}
+
 // makeLabelsForDruid returns the labels for selecting the resources
 // belonging to the given druid CR name.
 func makeLabelsForDruid(name string) map[string]string {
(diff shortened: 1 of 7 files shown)


AdheipSingh:
@himanshug any thoughts on this ?

AdheipSingh:
@himanshug in case you missed this :)

himanshug:
+1

AdheipSingh:
@himanshug thank you

AdheipSingh:
Open for initial reviews, no more WIP. Have added the interfaces and refactored as discussed, open for changes improvements :) 
@himanshug

controllers/druid/interface.go
```diff
@@ -0,0 +1,162 @@
...
+	List(ctx context.Context, sdk client.Client, drd *v1alpha1.Druid, selectorLabels map[string]string, emptyListObjFn func() runtime.Object, ListObjFn func(obj runtime.Object) []object) ([]object, error)
+	Get(ctx context.Context, sdk client.Client, nodeSpecUniqueStr string, drd *v1alpha1.Druid, emptyObjFn func() object) (object, error)
+}
+
+// Writer Interface
+type Writer interface {
+	Delete(ctx context.Context, sdk client.Client, drd *v1alpha1.Druid, obj runtime.Object, deleteOptions ...client.DeleteOption) error
+	Create(ctx context.Context, sdk client.Client, drd *v1alpha1.Druid, obj object) (string, error)
+	Update(ctx context.Context, sdk client.Client, drd *v1alpha1.Druid, obj object) (string, error)
+	StatusPatch(ctx context.Context, sdk client.Client, drd *v1alpha1.Druid, obj object, patch client.Patch) error
```
AdheipSingh, line 47: Instead of having seperate Patch for status, can have single method here. Args can be passed in case status needs to be matched else patch the object. Args can be boolean type

controllers/druid/handler.go
```diff
@@ -705,21 +644,16 @@ func sdkCreateOrUpdateAsNeeded(
 		if err := sdk.Get(context.TODO(), *namespacedName(obj.GetName(), obj.GetNamespace()), prevObj); err != nil {
```
AdheipSingh, line 644: Reason for not having called the get interfaces is cause GET shall log when it does not find the object, here the case is such that in case object not found create it and then log the error. So i ignored adding the writer.Get().

AdheipSingh:
@himanshug

controllers/druid/handler.go
```diff
@@ -592,49 +533,45 @@ func execCheckCrashStatus(sdk client.Client, nodeSpec *v1alpha1.DruidNodeSpec, m
...
-	pod := make([]*v1.Pod, 0)
-	for i := range podList.Items {
-		pod = append(pod, &podList.Items[i])
+	podList, err := readers.List(context.TODO(), sdk, drd, makeLabelsForDruid(drd.Name), func() runtime.Object { return makePodList() }, func(listObj runtime.Object) []object {
+		items := listObj.(*v1.PodList).Items
+		result := make([]object, len(items))
+		for i := 0; i < len(items); i++ {
+			result[i] = &items[i]
+		}
+		return result
```
nishantmonu51, line 544: I see similar logic iterated for almost every kind of object statefulset, PVC, podlist? Any way this can be refactored?

nishantmonu51 (approved):
mostly a refactoring patch, functionally there seems to be no change. 
👍

himanshug:
can you resolve the conflict ?

controllers/druid/interface.go
```diff
@@ -0,0 +1,172 @@
...
+type Writer interface {
+	Delete(ctx context.Context, sdk client.Client, drd *v1alpha1.Druid, obj runtime.Object, deleteOptions ...client.DeleteOption) error
+	Create(ctx context.Context, sdk client.Client, drd *v1alpha1.Druid, obj object) (string, error)
+	Update(ctx context.Context, sdk client.Client, drd *v1alpha1.Druid, obj object) (string, error)
+	Patch(ctx context.Context, sdk client.Client, drd *v1alpha1.Druid, obj object, status bool, patch client.Patch) error
+}
+
+// WriterFuncs struct
+type WriterFuncs struct {
+	deleteFunc func(ctx context.Context, sdk client.Client, drd *v1alpha1.Druid, obj runtime.Object) error
```
egor-ryashin, line 52: Function fields used when custom functions are assigned to them.
To implement an interface they aren't required.

This is an example when function fields are used https://www.geeksforgeeks.org/function-as-a-field-in-golang-structure/

controllers/druid/interface.go
```diff
@@ -0,0 +1,172 @@
...
+	createFunc func(ctx context.Context, sdk client.Client, drd *v1alpha1.Druid, obj object) (string, error)
+	updateFunc func(ctx context.Context, sdk client.Client, drd *v1alpha1.Druid, obj object) (string, error)
+	patchFunc  func(ctx context.Context, sdk client.Client, drd *v1alpha1.Druid, obj object, status bool, patch client.Patch) error
+}
+
+// ReaderFuncs struct
+type ReaderFuncs struct {
+	listFunc func(ctx context.Context, sdk client.Client, drd *v1alpha1.Druid, selectorLabels map[string]string, emptyListObjFn func() runtime.Object, ListObjFn func(obj runtime.Object) []object) ([]object, error)
+	getFunc  func(ctx context.Context, sdk client.Client, nodeSpecUniqueStr string, drd *v1alpha1.Druid, emptyObjFn func() object) (object, error)
```
egor-ryashin, line 61: The same here, I see they aren't assigned anywhere.

controllers/druid/interface.go
AdheipSingh: thank you for reviewing this @egor-ryashin 
 i missed this somehow, yes i can remove this, we can just use an empty struct.

controllers/druid/interface.go
AdheipSingh: agreed

controllers/druid/interface.go
himanshug: yeah seems legit to remove those.

controllers/druid/interface.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/druid-io/druid-operator/apis/druid/v1alpha1"
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	resourceCreated = "CREATED"
+	resourceUpdated = "UPDATED"
```
AdheipSingh, line 16: this can be typed too, just for consistency since the latter const declared are typed

controllers/druid/interface.go
```diff
@@ -0,0 +1,165 @@
...
+	resourceCreated DruidNodeStatus = "CREATED"
+	resourceUpdated DruidNodeStatus = "UPDATED"
+)
+
+const (
+	DruidNodeUpdateFail            string = "UPDATE_FAIL"
+	DruidNodeUpdateSuccess         string = "UPDATE_SUCCESS"
+	DruidNodeRollingDeploymentWait string = "ROLLING_DEPLOYMENT_WAIT"
+	DruidNodeDeleteFail            string = "DELETE_FAIL"
+	DruidNodeDeleteSuccess         string = "SUCCESS_FAIL"
```
AdheipSingh, line 26: This should be ```DELETE_SUCCESS```

AdheipSingh:
@himanshug 
I plan to update controller-runtime version, can we have this reviewed, would be easier with interfaces :)

PR #152, opened 2021-03-25
Fix CRD to support PVC name and labels. Also adding annotations to services.

### LINKED ISSUE #135: Wrong pvc for middle manager
create Pod druid-tiny-cluster-middlemanagers-0 in StatefulSet druid-tiny-cluster-middlemanagers failed error: failed to create PVC -druid-tiny-cluster-middlemanagers-0: PersistentVolumeClaim "-druid-tiny-cluster-middlemanagers-0" is invalid: metadata.name: Invalid value: "-druid-tiny-cluster-middlemanagers-0": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')

lum-splunk:
Fixes #135 

### Description

This PR should resolve the issue mentioned in #135. The PVC name will get prepended to the naming scheme, so that we don't end up with PVC names like "-druid-druid-historicals-0", which are invalid.

This PR also adds the ability to add annotations to services. My main rationale for this was to allow for the creation of internal AWS ELBs with annotations such as:

```
    service.beta.kubernetes.io/aws-load-balancer-healthcheck-interval: "5"

    service.beta.kubernetes.io/aws-load-balancer-healthcheck-timeout: "3"

    service.beta.kubernetes.io/aws-load-balancer-healthcheck-unhealthy-threshold: "2"

    service.beta.kubernetes.io/aws-load-balancer-internal: 0.0.0.0/0
```

<hr>

This PR has:

<hr>

 * `druid.apache.org_druids.yaml`

### DIFF
--- a/deploy/crds/druid.apache.org_druids.yaml
+++ b/deploy/crds/druid.apache.org_druids.yaml
@@ -3363,6 +3363,20 @@ spec:
                           metadata:
                             description: 'Standard object''s metadata. More info:
                               https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata'
+                            properties:
+                              annotations:
+                                additionalProperties:
+                                  type: string
+                                descriptions: 'Optional: Service annotations.'
+                                type: object
+                              name:
+                                description: 'Name of service.'
+                                type: string
+                              labels:
+                                additionalProperties:
+                                  type: string
+                                descriptions: 'Optional: Service labels.'
+                                type: object
                             type: object
                           spec:
                             description: Spec defines the behavior of a service. https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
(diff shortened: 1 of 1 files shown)


AdheipSingh:
Great !! Ill test it out, i have been using without openAPI for a while so dint come across this issue. 
Will check out with this CRD. 

cc @himanshug

dsx:
File `chart/templates/crds/druid.apache.org_druids.yaml` needs to be updated too.

AdheipSingh:
Plus once updated in helm , we also need to bump chart version.

dsx:
And not forget about helm annotation to leave CRD upon uninstall by default. I thinks at this point it would be safer to employ CI to do that, otherwise things will be forgotten.

himanshug:
thanks @lum-splunk , but unfortunately this doesn't work because CRD isn't hand written but generated. this will get overridden immediately.... CRD generation is controlled via comments added to https://github.com/druid-io/druid-operator/blob/master/apis/druid/v1alpha1/druid_types.go , so right markers would need to be added there so that generated CRD matches what is proposed here.

AdheipSingh:
agree with @himanshug 
make run will omit the changes anyways, so we need to add comments in types.go, code-gen will use those for crd generation

lum-splunk:
@himanshug @AdheipSingh  ok, I can make the updates as requested.

After making changes locally, how can I generate the CRD to test its working correctly?

himanshug:
@lum-splunk `make manifests` would re-generate the CRD.

y25ean:
I am currently experiencing this issue, when do you think that this pull request will be merged?

PR #191, opened 2021-05-22
Force cluster creation parallel, rolling deploy sequential

### LINKED ISSUE #177: Cluster Creation should be parallel, Rolling updates should be sequential
- When we have rolling deploy  set to true, upgrades are sequential. 
- When we have rolling deploy set to true, cluster creation is also sequential.
-------------------------------------------------------------------------------------------------------------------------------------------------------------
- A cluster can take upto 3-4 min of creation in a prod env. Which is slow, what we can do here is to make sure cluster creation is parallel regardless of rolling deploy set to true or false. 
-------------------------------------------------------------------------------------------------------------------------------------------------------------
- if ```sdkCreateOrUpdate()``` returns a ```CREATE``` string, we can ignore if ```ObjIsFullyDeployed()``` func,
- if ```sdkCreateOrUpdate()``` returns a ```UPDATE``` string, we can call if ```ObjIsFullyDeployed()``` func

@himanshug any thoughts on this ?

AdheipSingh:
Fixes #177.

- regardless of roling deploy set to true/false, cluster creation should be parallel.
- call objectFullyDeployed only when generation tag is > 1.
- generation tag increments on each apply/edit to the CR.
- During the first iteration generation tag is 1. 

### Description

<hr>

This PR has:

<hr>

 * `handler.go`

### DIFF
--- a/controllers/druid/handler.go
+++ b/controllers/druid/handler.go
@@ -179,7 +179,7 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 				}
 
 				// Ignore isObjFullyDeployed() for the first iteration ie cluster creation
-				// will force cluster cretion in parallel, post first iternation rolling updates
+				// will force cluster creation in parallel, post first iteration rolling updates
 				// will be sequential.
 				if m.Generation > 1 {
 					// Check Deployment rolling update status, if in-progress then stop here
@@ -209,7 +209,7 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 				execCheckCrashStatus(sdk, &nodeSpec, m)
 
 				// Ignore isObjFullyDeployed() for the first iteration ie cluster creation
-				// will force cluster cretion in parallel, post first iternation rolling updates
+				// will force cluster creation in parallel, post first iteration rolling updates
 				// will be sequential.
 				if m.Generation > 1 {
 					//Check StatefulSet rolling update status, if in-progress then stop here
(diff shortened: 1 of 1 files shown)


controllers/druid/handler.go
```diff
@@ -203,10 +208,15 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 				// Default is set to true
 				execCheckCrashStatus(sdk, &nodeSpec, m)
 
-				//Check StatefulSet rolling update status, if in-progress then stop here
-				done, err := isObjFullyDeployed(sdk, nodeSpecUniqueStr, m, func() object { return makeStatefulSetEmptyObj() })
-				if !done {
-					return err
+				// Ignore isObjFullyDeployed() for the first iteration ie cluster creation
+				// will force cluster cretion in parallel, post first iternation rolling updates
```
AdheipSingh, line 212: nit: ```creation```

controllers/druid/handler.go
```diff
@@ -178,10 +178,15 @@ func deployDruidCluster(sdk client.Client, m *v1alpha1.Druid) error {
 					return nil
 				}
 
-				// Check Deployment rolling update status, if in-progress then stop here
-				done, err := isObjFullyDeployed(sdk, nodeSpecUniqueStr, m, func() object { return makeDeploymentEmptyObj() })
-				if !done {
-					return err
+				// Ignore isObjFullyDeployed() for the first iteration ie cluster creation
+				// will force cluster cretion in parallel, post first iternation rolling updates
```
AdheipSingh, line 182: nit: ```creation```

PR #200, opened 2021-05-27
refactor objfullydeployed()

### LINKED ISSUE #196: refactor objFullyDeployed()
- While downscaling some druid nodes i see, the sts revision numbers remain same and the operator sees the sts object as fully deployed,  which is not the case since replica count as desired and current still differ.
- obj should be fully deployed if replicas state is also met, operator needs to have additional checks too.
- Deployment object can have the same replicas but conditions might differ, need to have a check on conditions  too.

will send a PR for this

AdheipSingh:
Fixes #196

### Description
- main goal was to detect failures during rolling deploy.
- Deployment has conditions to detect so have added those.
- Statefulsets does not have status conditions its TODO in upstream https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/apps/types.go#L217

<hr>

This PR has:

<hr>

 * `handler.go`

### DIFF
--- a/controllers/druid/handler.go
+++ b/controllers/druid/handler.go
@@ -5,6 +5,7 @@ import (
 	"crypto/sha1"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"reflect"
 	"regexp"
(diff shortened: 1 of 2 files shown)


nishantmonu51 (approved):
👍

PR #223, opened 2021-07-02
Introduce State Driven Status, use record pkg for event and clean logging

### LINKED ISSUE #190: Druid Node Status, Druid Cluster Status and Events
## ISSUE

- Currently the druid operator sends events on various state changes and actions it takes in order to meet the desired state of the CR. 
- The events emitted end up getting repeated and pollute the kubectl describe and get events cli calls.
- Ex: In case rolling update is ongoing, operator keeps on emitting events till the time rolling get deploy completes. This ends up with ROLLING_DEPLOY_WAIT event all over the describe of CR. This also pollutes kubectl get events :| 
- On further debug, as an operator i would like to get events only when a state change has occurred, if the previous state has not changed the operator should not send the event. This behaviour was seen in deployment controller of kubernetes where events are only emitted on state change.
- so the goal here is too make sure events are emitted only on state change, in case the previous state is same do not emit event. 

------------------------------------------------------------------------------------------------------------------------------------------------------------------------

## Approach

- After going through various approaches here's the approach i finalise, most of this is taken from the current controller implementations in upstream kubernetes.

- The first thing is here, we need to have an improved status of the Druid CR, which shall include 
1. DruidNodeTypeStatus - represents the state of each nodeType 
2. DruidClusterFinalStatus -  represents the cumulative state of all the nodeType

- Their is no way straight way to get druid final status, the operator runs a for loop reconciling each nodeType, at that particular moment we are aware of the current status of the particular nodeType.
- Ex: In case i trigger a rolling update which results in a cold nodeType crashing and other node types are still running, the DruidClusterFinalStatus should be NOT_READY at the same time DruidNodeTypeStatus should show the cold is NOT_READY. Event emitted should change from ROLLING_DEPLOYMENT_WAIT to COLD_NODE_CRASHING. Till the time the cold is not brought up the operator should not emit event.

- Events should check the state from status of the druid CR and get emitted, if prev state is still in-progress dont emit event.
- Druid Node Transition shall be displayed in the druid CR.
- On the basis of Druid Node Transitions we build the druid cluster final status.

### LINKED ISSUE #195: Operator Events not using client go record pkg ?
@himanshug i have few questions around the current implementations of the sendEvents()

- since i was spending some time in trying to fix the events polluting the describe of druid, i came across a standard implementations of events in k8s controllers.
1. DruidReconciler struct does not have any fields related to events records.
2. Any specific reason, you planned to keep events out of this struct. and not use pkg "k8s.io/client-go/tools/record" ? 
example:
```
// DruidReconciler reconciles a Druid object
type DruidReconciler struct {
	client.Client
	Log    logr.Logger
	Scheme *runtime.Scheme
	Record record.EventRecorder
	// reconcile time duration, defaults to 10s
	ReconcileWait time.Duration
}
```
- I see records pkg has a sink interface, where we can patch and update events rather than creating them each time. 
https://github.com/kubernetes/client-go/blob/master/tools/record/event.go#L47

- I also think since this Reconcile method implements the DruidReconciler, not sure how will deployDruid(), will use ```r.Record```, might need to create a custom interface as per druid operator use case. Just saying though :)

AdheipSingh:
Fixes #195 
Fixes #190

#### Changes 
1. Introduce druid cluster status
2. Use event recorder pkg 
3. clean up unnecessary logging
4. zero pollution of events as well as logs, reduces the requests on k8s api
5. state driven status.
6. EventEmitter interface - can be extended in the future to emit events to various backends ( kafka, statsd just like druid does )

#### Logging

- A delete, and update logs on deletion of CR ( finalizer )
image

<hr>

This PR has:

<hr>

 * `handler.go`
 * `types.go`
 * `status.go`
 * `handler.go`


controllers/druid/status.go
```diff
@@ -0,0 +1,86 @@
...
+
+	var reason string
+
+	if nodeCondition == v1alpha1.DruidClusterReady {
+		nodeTierOrType = ""
+		reason = "All Druid Nodes are in Ready Condition"
+	} else if nodeCondition == v1alpha1.DruidNodeRollingUpdate {
+		reason = "Druid Node [" + nodeTierOrType + "] is Rolling Update"
+	} else if err != nil {
+		nodeTierOrType = ""
```
AdheipSingh, line 31: this can kept to "All"

controllers/druid/interface.go
```diff
@@ -75,105 +91,188 @@ type objectList interface {
...
+	}
+}
+
+// EmitEventOnDelete shall emit event on DELETE operation
+func (e EmitEventFuncs) EmitEventOnDelete(obj, deleteObj object, err error) {
+	if err != nil {
+		errMsg := fmt.Errorf("Error deleting object [%s] in namespace [%s] due to [%s]", deleteObj.GetObjectKind().GroupVersionKind().Kind, deleteObj.GetNamespace(), err.Error())
+		e.Event(obj, v1.EventTypeWarning, string(druidNodeDeleteFail), errMsg.Error())
+	} else {
+		msg := fmt.Sprintf("Successfully deleted object [%s] in namespace [%s]", deleteObj.GetObjectKind().GroupVersionKind().Kind, deleteObj.GetNamespace())
```
AdheipSingh, line 253: add ```object.GetName()``` on deletion msg

AdheipSingh:
@himanshug will need your review here.

himanshug:
sorry, will review it this week.

AdheipSingh:
@himanshug :)

himanshug:
I took a quick look today, LGTM

We can merge this if you have already had this deployed on some clusters and it is functioning as expected.

AdheipSingh:
> We can merge this if you have already had this deployed on some clusters and it is functioning as expected.

yes, we have been running this on prod for couple of months.

@nishantmonu51 waiting on you for this.

nishantmonu51 (approved):
👍

Issue #225, opened 2021-07-06
Restrict K8s role definitions to actually needed roles


DipeshCS:
We are planning to use Druid operator in Production at Zalando and we realized that [k8s role definition](https://github.com/druid-io/druid-operator/blob/master/deploy/role.yaml) is too generous giving access to lot more resources for operator than needed.  

Few examples would be:
- Usage of daemonsets is missing
- Usage of servicemonitors is missing
- Few resources use `*` but it is good practice to explicitly list them out
and many more.

It would be great if it can be fixed and made production-ready. 
Let me know if you need more information. 

Thanks

AdheipSingh:
@DipeshCS 
agree with your observations and points listed out. Do you mind sending a patch and we can have it reviewed :)
Thanks

PR #226, opened 2021-07-12
Restrict K8s role definitions to actually needed roles

piotrkmita:
Fixes [#255](https://github.com/druid-io/druid-operator/issues/225).

- daemonsets, replicasets and servicemonitors, finalizers removed - usage of them are missing
- extensions/ingress removed - it has been removed for a while (https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/)
- rest of resources changed from `*` to an explicit list.
<hr>

This PR has:
- [X] been tested for backward compatibility on a real K*S cluster by applying the changes introduced here on an existing Druid cluster. If there are any backward incompatible changes then they have been noted in the PR description.

<hr>

 * `role.yaml`

### DIFF
--- a/chart/templates/role.yaml
+++ b/chart/templates/role.yaml
@@ -32,6 +32,7 @@ rules:
       - events
     verbs:
       - create
+      - patch
   - apiGroups:
       - apps
     resources:
--- a/chart/templates/role.yaml
+++ b/chart/templates/role.yaml
@@ -13,78 +13,82 @@ metadata:
   labels:
     {{- include "druid-operator.labels" . | nindent 4 }}
 rules:
-- apiGroups:
-  - ""
-  resources:
-  - pods
-  - services
-  - services/finalizers
-  - endpoints
-  - persistentvolumeclaims
-  - events
-  - configmaps
-  - secrets
-  verbs:
-  - '*'
-- apiGroups:
-  - apps
-  resources:
-  - deployments
-  - daemonsets
-  - replicasets
-  - statefulsets
-  verbs:
-  - '*'
-- apiGroups:
-    - policy
-  resources:
-    - poddisruptionbudgets
-  verbs:
-    - '*'
-- apiGroups:
-    - autoscaling
-  resources:
-    - horizontalpodautoscalers
-  verbs:
-    - '*'
-- apiGroups:
-    - extensions
-    - networking.k8s.io
-  resources:
-    - ingresses
-  verbs:
-    - '*'
-- apiGroups:
-  - monitoring.coreos.com
-  resources:
-  - servicemonitors
-  verbs:
-  - get
-  - create
-- apiGroups:
-  - apps
-  resourceNames:
-  - druid-operator
-  resources:
-  - deployments/finalizers
-  verbs:
-  - update
-- apiGroups:
-  - ""
-  resources:
-  - pods
-  verbs:
-  - get
-- apiGroups:
-  - apps
-  resources:
-  - replicasets
-  verbs:
-  - get
-- apiGroups:
-  - druid.apache.org
-  resources:
-  - '*'
-  verbs:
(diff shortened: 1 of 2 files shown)


deploy/role.yaml
```diff
@@ -4,77 +4,81 @@ metadata:
...
+      - patch
+      - delete
+  - apiGroups:
+      - druid.apache.org
+    resources:
+      - druids/status
+    verbs:
+      - get
+      - update
+      - patch
```
AdheipSingh, line 84: nit: format

AdheipSingh:
@piotrkmita thanks for the PR. 

> daemonsets, replicasets and servicemonitors, finalizers removed - usage of them are missing

druid operator has a finalizer implementation, did you test the PR with the default behaviour of the finalizer ( enabled by default ) ? 
BTW what version of k8s was this PR testing on ?

Thanks

piotrkmita:
Hey @AdheipSingh,

Our k8s team is not allowing setting finalizers on built-in types. This is the response we've got from them.
> The finalizers should only be on the custom resource. I don't think you can even set finalizers on built-in types, so it does not make sense to allow it in RBAC, which is why we don't allow this.

That's why we've tried default behavior which worked fine and in the operator logs we could see messages from finalizer
image
We've also tried using `disablePVCDeletionFinalizer` parameter to make sure it's working properly.

We’ve tested it with both minikube and our test cluster. Here are the versions:
minikube:
`Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.2", GitCommit:"092fbfbf53427de67cac1e9fa54aaa09a28371d7", GitTreeState:"clean", BuildDate:"2021-06-16T12:53:14Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"linux/amd64”}`

test cluster:
`Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.8", GitCommit:"5575935422cc1cf5169dfc8847cb587aa47bac5a", GitTreeState:"clean", BuildDate:"2021-06-16T12:53:07Z", GoVersion:"go1.15.13", Compiler:"gc", Platform:"linux/amd64"}`

Let me know if we missed something there,
Thanks!

deploy/role.yaml
piotrkmita: Just to make sure, is it about newline missing?

AdheipSingh: yes :)

AdheipSingh:
@piotrkmita 
thanks for the info. 
LGTM , though ill run a test on in house cluster and revert back.

AdheipSingh (approved):

nishantmonu51 (approved):
👍

chart/templates/role.yaml
```diff
@@ -13,78 +13,82 @@ metadata:
...
+      - watch
+      - create
+      - update
+      - delete
+  - apiGroups:
+      - ""
+    resources:
+      - events
+    verbs:
+      - create
```
AdheipSingh, line 34: add ```patch``` too for events. 
this will be needed for this PR  https://github.com/druid-io/druid-operator/pull/223
currently operator's creates events each time, with this https://github.com/druid-io/druid-operator/pull/223 a re-occurring event will be patched.

AdheipSingh (changes requested):

AdheipSingh:
have tested this PR on v1.21.2 do and 1.19 gke.
LGTM
Thanks @piotrkmita 

@himanshug @nishantmonu51 good to go :)

AdheipSingh (approved):

nishantmonu51 (approved):
👍

AdheipSingh:
@himanshug waiting for your approval :)

himanshug:
thanks, took a cursory look, LGTM

PR #268, opened 2022-01-07
Add command args to container creation (#6)

CodingParsley:
Fixes #XXXX.

### Description

<hr>

This PR has:

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`


AdheipSingh (changes requested):

apis/druid/v1alpha1/druid_types.go
```diff
@@ -53,9 +53,16 @@ type DruidSpec struct {
...
+	// Required: Command to be run on container start
 	StartScript string `json:"startScript"`
 
+	// Optional: bash/sh entry arg. Set startScript to `sh` or `bash` to customize entryArg
+	// For example, the container can run `sh -c "${EntryArg} && ${DruidScript} {nodeType}"`
+	EntryArg string `json:"entryArg,omitempty"`
+
+	// Optional: Customized druid shell script path. If not set, the default would be "bin/run-druid.sh"
+	DruidScript string `json:"druidScript,omitempty"`
+
```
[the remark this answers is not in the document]
AdheipSingh: @CodingParsley :)

controllers/druid/testdata/druid-test-cr-start-command.yaml
```diff
@@ -0,0 +1,369 @@
+apiVersion: "druid.apache.org/v1alpha1"
+kind: "Druid"
+metadata:
+  name: druid-test
+  namespace: test-namespace
+spec:
+  image: himanshu01/druid:druid-0.12.0-1
+  startScript: sh
+  entryArg: echo 'Hello World'
+  DruidScript: druid-test.sh
```
AdheipSingh, line 10:
```suggestion
  druidScript: druid-test.sh
```

AdheipSingh (changes requested):

CodingParsley:
@AdheipSingh Thank you ❤️

AdheipSingh:
> @AdheipSingh Thank you ❤️

ill run this locally and test, cc @nishantmonu51 for approval to run CI.

just a question, does this PR work if you specify image for all nodeSpecs and also for individual nodeSpec ?

AdheipSingh:
@xvrl can you approve this PR to run the CI. :)

CodingParsley:
> just a question, does this PR work if you specify image for all nodeSpecs and also for individual nodeSpec ?

Yes, this PR does not change the image of the container

apis/druid/v1alpha1/druid_types.go
```diff
@@ -91,9 +91,16 @@ type DruidSpec struct {
 	// +optional
 	DeleteOrphanPvc bool `json:"deleteOrphanPvc"`
 
-	// Required: path to druid start script to be run on container start
+	// Required: Command to be run on container start
 	StartScript string `json:"startScript"`
 
+	// Optional: bash/sh entry arg. Set startScript to `sh` or `bash` to customize entryArg
+	// For example, the container can run `sh -c "${EntryArg} && ${DruidScript} {nodeType}"`
```
xvrl, line 98: I don't think this is fully backwards compatible. Since we are now wrapping this in a shell, any signals sent to the process running the container will be sent to the shell and not to the JVM. The druid startup script ensures that the JVM will replace the shell, but in this case I do not think that will happen.

Is there a reason why the changes cannot be done inside of the druid startup script? It might be useful to provide an example use-case to clarify that.

CodingParsley: It just gives the user an option to run druid startup script in shell, but the default is still 
https://github.com/druid-io/druid-operator/pull/268/files#diff-6b7cea906298d4262b867f6e81f41caa89c2daeb67247bd95169e93ed80a00b4R1146

xvrl: understood, but having it behave differently is not a good idea, and the nuances of using one or the other are too subtle in my opinion.

CodingParsley: Its true the same can be achieved by adapting `druid.sh`. Adding custom arguments just provides a more lightweight way to update the container startup command than updating `druid.sh` and building a new image.

xvrl: understood, but can we include a concrete use-case in the description that would warrant this feature specifically in operator? it would help understand the motivation for this change.

xvrl: wouldn't it make more sense to let someone override the entry point in the helm chart rather than having to customize it in the operator config?

CodingParsley:
@xvrl Maybe another look? 🙏

controllers/druid/handler.go
```diff
@@ -1308,7 +1324,8 @@ func makePodSpec(nodeSpec *v1alpha1.DruidNodeSpec, m *v1alpha1.Druid, nodeSpecUn
 		v1.Container{
 			Image:           firstNonEmptyStr(nodeSpec.Image, m.Spec.Image),
 			Name:            fmt.Sprintf("%s", nodeSpecUniqueStr),
-			Command:         []string{firstNonEmptyStr(m.Spec.StartScript, "bin/run-druid.sh"), nodeSpec.NodeType},
+			Command:         getCommand(nodeSpec, m),
+			Args:            getEntryArg(nodeSpec, m),
```
xvrl, line 1328: instead of adding this complex custom logic, would it make more sense to let the user explicitly override `Command` and `Args` directly?

AdheipSingh: In case of an override for a ```Command``` or an ```Arg``` how can it be done directly ? Override in the CR ? or some way of templating to override it out

xvrl: yes, I meant to let the user override those in the CR, rather than creating our own special variables

xvrl: Right now we have `startScript`, `druidScript`, and `entryArg`. The way these get combined is quite confusing, and it's not clear why we need 3.

If I specify `druidScript`, but neither `startScript` nor `entryArg`, then `druidScript` is ignored and we run the default `bin/run-druid.sh` so `druidScript` doesn't seem to control the Druid script in all cases unlike what intuition might suggest

This is quite confusing logic for someone new to this (especially since there is very little documentation).

CodingParsley: I tried to do that initially. The problem is that, we have to add the node type after the druid.sh script for each node dynamically, otherewise the user needs to specify separate container commands and arguments for each node. 
Therefore, we can not just let them overwrite the commands & args freely

AdheipSingh: > This is quite confusing logic for someone new to this (especially since there is very little documentation).

Yeah it can be confusing. I am not a user of this feature/requirement so don't have any comments on the use case.
Regardless i leave it for your team, i personally feel that configuration management is a total abstraction, operator CR should be a minimal viable spec to support what is needed to support overrides, templating etc. :) 

Also @CodingParsley how about just adding StartScript to each nodeSpec also, currently is just present on the cluster scope ( its common to all nodesSpecs ). 
https://github.com/druid-io/druid-operator/blob/master/apis/druid/v1alpha1/druid_types.go#L95

CodingParsley: > just adding StartScript to each nodeSpec also, currently is just present on the cluster scope ( its common to all nodesSpecs )
Sounds like a solution! Will rewrite the PR

PR #283, opened 2022-04-08
allow scale down of druid pods to zero

### LINKED ISSUE #99: operator should allow scale replicas to zero
- validation , has replicas < 1, which leads not to scale to zero.
- incase ingestion/querying not happening MM/brokers or even hot/cold can be scaled to zero.
- i guess we can remove replicas from validation ?

AdheipSingh:
Fixes #99 .

### Description

<hr>

This PR has:

<hr>

 * `handler.go`


nishantmonu51 (approved):
👍

PR #286, opened 2022-04-11
suggested update

mikkeschiren:
We had a hard time getting the Helm chart to work, and these are the fixes we did to get it to work, I don't know if this contribution is valid or not, but this worked well for us, so wanted to share it.


mikkeschiren:
Scratch this - this was in the wrong place, sorry.

PR #309, opened 2022-09-29
multiple watched namespaces

### LINKED ISSUE #308: Helm chart templates for roles and service accounts are not correct
There are a few problems there in the templates.
1) Doc says the operator can watch multiple namespaces, by setting `env. WATCH_NAMESPACE=NS1,NS2`. But the [templates](https://github.com/druid-io/druid-operator/blob/5abef01594ff953ba8a092aa7d392c06d682b167/chart/templates/role.yaml#L3-L9) do not handle multiple namespaces.
2) Permissions on `storage.k8s.io` is needed, otherwise `scalePvcSts` won't work

zhangluva:
Fixes #308 

### Description
* Fixes problem when multiple namespaces are watched
* Fixes problem when `scaleStsPvc` is set to `true`\

When multiple namespaces are watched, configured by set `env.WATCH_NAMESPACE=ns1,ns2`, a role needs to be created in each of the namespaces.
Also, a clusterrole with list/get permission on `storage.k8s.io` api is needed in order for `scaleStsPvc` to work. All roles need to bind to the operator's service account.

<hr>

This PR has:
- [X] been tested for backward compatibility on a real K*S cluster by applying the changes introduced here on an existing Druid cluster. If there are any backward incompatible changes then they have been noted in the PR description.

<hr>

 * `chart/template/role.yaml`
 * `chart/templates/role_binding.yaml`
 * `chart/templates/watched_namespace.yaml`

### DIFF
--- a/chart/templates/role.yaml
+++ b/chart/templates/role.yaml
@@ -1,17 +1,17 @@
 {{- if .Values.rbac.create }}
+{{- $env := .Values.env }}
+{{- if and ($env.WATCH_NAMESPACE) (ne $env.WATCH_NAMESPACE "default") }}
+# Split WATCH_NAMESPACE by commas and loop on them
+{{- $watchedNamespaces := (split "," $env.WATCH_NAMESPACE) -}}
+{{- range $watchedNamespaces }}
+---
 apiVersion: rbac.authorization.k8s.io/v1
-{{- if .Values.env.WATCH_NAMESPACE }}
 kind: Role
-{{- else }}
-kind: ClusterRole
-{{- end }}
 metadata:
-{{- if .Values.env.WATCH_NAMESPACE }}
-  namespace: {{ .Values.env.WATCH_NAMESPACE }}
-{{- end }}
-  name: {{ template "druid-operator.fullname" . }}
+  namespace: {{ . }}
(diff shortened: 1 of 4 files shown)


chart/templates/role.yaml
```diff
@@ -94,3 +183,22 @@ rules:
...
+    {{- include "druid-operator.labels" $ | nindent 4 }}
+rules:
+  - apiGroups:
+      - storage.k8s.io
+    resources:
+      - storageclasses
+    verbs:
+      - get
+      - list
+      - watch
```
AdheipSingh, line 204: @zhangluva on which cloud platform was this needed, what was the k8s version ?

AdheipSingh (commented):
thanks for the pr, will check it out.

chart/templates/role.yaml
zhangluva: Tested on EKS, `v1.22.11-eks-18ef993`

AdheipSingh:
@nishantmonu51 kindly approve to run CI

chart/templates/role.yaml
AdheipSingh: @zhangluva just the indentation. Rest LGTM Thanks

AdheipSingh (approved):

PR #284, opened 2022-04-08
make nodeselector consistent

AdheipSingh:
Fixes #55.

### Description

<hr>

This PR has:
- [ ] been tested on a real K8S cluster to ensure creation of a brand new Druid cluster works.
- [ ] been tested for backward compatibility on a real K*S cluster by applying the changes introduced here on an existing Druid cluster. If there are any backward incompatible changes then they have been noted in the PR description.
- [ ] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
- [ ] added documentation for new or modified features or behaviors.

<hr>

##### Key changed/added files in this PR
 * `types.go`
 * `handler.go`
 * `deepcopy.go`

PR #328, opened 2022-12-19
Nodespec "nodeSelector" should override Clusterspec

CodingParsley:
Fixes https://github.com/druid-io/druid-operator/pull/284

### Description
We have two nodeSelectors at spec level and node level. The expected behavior is that the node's nodeSelector should override the common spec's nodeSelector. So, changing the order.

<hr>

This PR has:

<hr>


AdheipSingh (approved):
LGTM thanks @CodingParsley 
@xvrl , kindly approve to run workflow + merge.

PR #347, opened 2024-06-04
Adding checks to ensure historical tiers also undergo rollingDeploy and do not deploy in parallel

aruraghuwanshi:
### Description

In the current Druid Operator, when rollingDeploy is enabled, the expectation is that the nodes will restart one at a time in the pre-defined order. In the case where we have multiple tiers within historicals, that is equivalent to having multiple Stateful Sets of NodeType historical. The Operator then does not stop to check whether each historical tier Statefulset is deployed and ends up deploying all historical tiers one after the other without waiting for a full deployment of the previous StatefulSet.

This PR aims to solve this issue by introducing a check on all historical tiers present in that cluster, if rollingDeploy is enabled, before going ahead with the next tier's deployment. In the specific case when we have replicas of datasource distributed across multiple tiers, we do not want all the tiers going down simultaneously (with rollingDeploy enabled), which might result in none of the segments being available, and hence downtime. This PR would solve that.

<hr>

This PR has:

<hr>

 * `MyFoo`
 * `OurBar`
 * `TheirBaz`

### DIFF
--- a/controllers/druid/handler.go
+++ b/controllers/druid/handler.go
@@ -41,6 +41,8 @@ const (
 	finalizerName                = "deletepvc.finalizers.druid.apache.org"
 )
 
+//Initialize the slice to store node-specific unique identity strings for historical tiers
+var historicalTierList []string
 var logger = logf.Log.WithName("druid_operator_handler")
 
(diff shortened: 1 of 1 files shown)


yurmix:
@aruraghuwanshi this repository is deprecated, see #336 
Project is now maintained in https://github.com/datainfrahq/druid-operator

